PR #3, opened 2015-04-21
fixing typos

dret:


hallvors:
Wow, thanks. Never noticed those :)

Issue #57, opened 2017-11-30
Should clipboard actions consider whether selection has focus?


xiaochengh:
Consider a page with

```
<div contenteditable>text</div>
<input type=checkbox>
```

And steps:
1. Copy something into clipboard
2. Select "text"
3. Click the checkbox to move focus. Now we have an unfocused selection on "text"
4a. Press Ctrl+C (or Cmd+C on Mac) to trigger a copy action
4b. Press Ctrl+X (or Cmd+X on Mac) to trigger a cut action
4c. Press Ctrl+V (or Cmd+V on Mac) to trigger a paste action

Should these actions have effects? And what should be the clipboard event targets?

The current spec says nothing about focus, so all actions should be effective and targetting DIV following the spec.

While copying without focus may be reasonable, cut and paste are somehow different because they modify the current document. And it doesn't seem quite desirable to change an unfocused part of document in a user action. [1]

Current behaviors of different browsers (demo at https://jsfiddle.net/54scvon9/5/):

Chrome 62.0.3202.94:
copy: effective, targetted on DIV
cut & paste: ineffective, targetted on DIV [2]

Firefox 57:
All effective and targetted on the text node

Safari 11.0.1(12604.3.5.1.1):
All effective and targetted on DIV

Edge 40.15063.0.0:
All effective and targetted on DIV

[1] Discussed in crbug.com/771604
[2] This seems to be a Chrome bug: crbug.com/755459

hugoholgersson:
A focused selection is always visible. In this case we have a visible but unfocused selection.

Users expect effective CTRL+C/V/P anytime a selection is _visible_? In this case, users can't know if the selection is focused or not. They see a blue selection so they expect interaction with it?

To fulfill that expectation, maybe CTRL+C/V/P should move focus to a visible&&unfocused selection? Hmm, could be complicated; the focus move would need to be done after any ClipboardEvent handler has had a chance to cancel the default operation => ClipboardEvents must target the visible selection.

Either way, I think it is safe to spec:
"Set target to be the element that contains the start of the visible selection in document order, or the body element if there is no visible selection or cursor"

xiaochengh:
I'm not sure if users expect effective cut/paste when there is unfocused visible selection. It can be confusing if cut/paste still modifies content of a contenteditable even though the user has explicitly moved the focus away.

[...]

PR #58, opened 2017-12-15
Emphasize visability in the "pick target"-algorithm

hugoholgersson:
This change addresses the ambiguity of unfocused and hidden
selections (#57) by stating that only a visible selection can be a
ClipboardEvent-target.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1323,17 +1323,9 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 
 		1. Let |target| be set as follows:
 
-			1. If the context is editable, then
-
-				1. Set |target| to be the element that contains the
-					start of the selection in document order, or
-					[=the body element=] if there is no selection or cursor.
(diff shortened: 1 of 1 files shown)


hugoholgersson:
Marked as non substantive for IPR from ash-nazg.

garykac:
Sorry for taking so long to get around to this. Can you make these edits to the `index.bs` file instead of the HTML file?
The .HTML file is auto-generated from the .BS file, so they'll be overwritten if we edit the HTML file directly.
(I've updated the readme.md file to note that contributions should be made to the .BS file -- it wasn't clear before).

hugoholgersson:
I've moved the changes to `index.bs`. PTAL. Please merge the changes if everything's all right.

garykac:
LGTM, although the fact that both part of the "if content is editable" are the same is actually a bug that I introduced when converting the spec.
I'll merge this and then re-add the IF along with the correct ELSE clause.

Issue #61, opened 2018-01-15
Define each event type is composed or not, update the spec


TakayoshiKochi:
It looks current ClipboardEvent doesn't define anything about whether each event is composed or not.

Once it was discussed at https://github.com/w3c/webcomponents/issues/513#issuecomment-224468080) but no one has ever revisited this.

/cc @annevk @hayatoito @rniwa

TakayoshiKochi:
The current status is that, as the spec says nothing about `composed` flag, but `ClipboardEventInit` dictionary inherits `EventInit`, `composed` flag is false by default.

Considering the similarity with UI events, shall we flip the default to be true?

hayatoito:
@TakayoshiKochi  What's your opinion?

TakayoshiKochi:
My opinion is composed=true by default.

annevk:
For these events, when they're not created synthetically? If so, seems fine to me and matches other UI events.

TakayoshiKochi:
@annevk yes, so follow other UI events.

I found @rniwa already commented: https://github.com/w3c/webcomponents/issues/513#issuecomment-230989199
> we should make `copy`, `paste`, `cut` and their `before*` variants composed since they don't have any node leakage problem, and yet are very important user gestures.

So I'll start preparing a PR for the clipboard spec.

TakayoshiKochi:
Did not find any MS people in the original thead https://github.com/w3c/webcomponents/issues/513, but cc @travisleithead - sorry I missed you.

hayatoito:
Could you help me to understand the motivation of making clipboard events composed one?

Given that users can't access a node where an event is fired, what is the use case for that?

TakayoshiKochi:
WebKit implements clipboard events as composed=true by default, while Blink implements them composed=false.  So we'd like to clarify which is the default.

I thought use cases are obvious - like catching the timing when copy or paste operation happens within shadow root, observe it from outside, and do something against it.  For example, if you want to use an editor component with shadow root, the embedder of the component may want to observe copy/paste events.  Even without looking at the real target, you can access ClipboardData via the event.

hayatoito:
Thanks, however, the explanation is still lacking why a light tree want to know copy & paste events which is happening in a shadow tree. What is the difference between clipboard events and other events?

I meant your explanation is generic and might be applied to all kinds of events which are happening in a shadow tree; "A light tree want to observe what is happening in a shadow tree".

[...]

PR #62, opened 2018-01-16
Make all clipboard events composed by default.

TakayoshiKochi:
This resolves https://github.com/w3c/clipboard-apis/issues/61

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -260,7 +260,7 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 			data will be copied to the [=system clipboard=].
 			The current document selection is not affected.
 
-			The [=copy=] event bubbles and is cancelable.
+			The [=copy=] event bubbles, is cancelable, and is composed.
 			
 			See [[#copy-action]] for a detailed description of the processing model for
 			this event.
@@ -284,7 +284,7 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 			be an empty list. Note that the [=cut=] event will still be fired
 			in this case.
 
-			The [=cut=] event bubbles and is cancelable.
+			The [=cut=] event bubbles, is cancelable, and is composed.
 
 			See [[#cut-action]] for a detailed description of the processing model for
 			this event.
@@ -306,7 +306,7 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 			The paste action has no effect in a non-[=editable context=],
 			but the [=paste=] event fires regardless.
 
-			The [=paste=] event bubbles and is cancelable.
+			The [=paste=] event bubbles, is cancelable, and is composed.
 
 			See [[#paste-action]] for a detailed description of the processing model for
 			this event.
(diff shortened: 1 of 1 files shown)


TakayoshiKochi:
Please hold on reviewing until #61 concludes.

garykac:
Can you make these edits to the `index.bs` file instead of the HTML file?
The .HTML file is auto-generated from the .BS file, so they'll be overwritten if we edit the HTML file directly.

TakayoshiKochi:
Done.  Could you take another look?

foolip:
Hi all,

This quarter @rakuco and I have been looking into a [normative spec change without tests](https://docs.google.com/document/d/1kFYTqUMEbo9p87i6wAzORoicMWgHsuYwL2CgtvdhDaE/edit#heading=h.gn90gr74sg9s) to help us understand how [testing in standards](https://github.com/foolip/testing-in-standards) is working out in practice, and where there's more infrastructure work needed to make it practical.

This week I ended up here, and I think https://github.com/w3c/clipboard-apis/pull/58 and https://github.com/w3c/clipboard-apis/commit/11619a7 (no PR?) might also be testable, but less obviously so.

It looks like there weren't any tests for any events at all in https://github.com/w3c/web-platform-tests/tree/master/clipboard-apis, presumably because they'd have to be manual. To convince myself it's possible, I wrote a manual test that exercises this change:
https://github.com/w3c/web-platform-tests/pull/9626 (review appreciated)

Question for everyone here: what automation bits would you need in https://w3c.github.io/webdriver/webdriver-spec.html and exposed in http://web-platform-tests.org/writing-tests/testdriver.html to be able to reliably test changes to this spec? If you could file an issue on the web-platform-tests repo and add the type:untestable label, that'd be much appreciated.

In the meantime, do you think that writing manual tests is worthwhile? Is there any useful advice that should go in https://github.com/w3c/clipboard-apis/blob/master/CONTRIBUTING.md#tests?

TakayoshiKochi:
Thanks @foolip for catching this.

As you imagined, this needs a manual test (could be automated by test driver?) and I was thinking of writing some tests while I'm changing the implementation in Blink - and you went ahead of me :)

I thought we can do document.execCommand('copy'), but turned out using execCommand() is not reliable :(

TakayoshiKochi:
FYI, this is my wip CL
https://chromium-review.googlesource.com/c/chromium/src/+/923671

foolip:
@kereliuk, do you have an update on testdriver.js support for keyboard input? I think that could be used to copy-paste, although it'd assume something about the key combinations that do it on various platforms. With that these tests should be possible to automate.

My follow-up question is whether it is generally true that changes in the Clipboard API spec get tests from implementations shortly thereafter? In that case it seems like a very good place to be. But I wonder how often things fall through the cracks, so that spec changes are made but no implementer ever follows, and it takes years before anyone notices?

kereliuk:
Working on the keyboard input/direct WebDriver now, should have a PR up soon.

There has been discussion about adding some sort of spec support for clipboard stuff in WebDriver like copy and paste and not just have it as using the keyboard, (notably at TPAC last year). But not much of a consensus was reached and it becomes tricky if we want to use the system clipboard especially in the case where there are multiple sessions running at the same time which is common.

foolip:
I guess we could put copy/paste/cut with keyboard in testdriver.js and implement using WebDriver, so that it can be overridden on platforms where the keyboard shortcuts aren't the same.

We really can't get around the fact that the clipboard is global I think, at most we could try to warn about it somewhere.

PR #82, opened 2018-09-21
Specify NotAllowedError when permission is not granted

### LINKED ISSUE #76: Clarify promise rejection value
I'm a bit unsure of what to expect back from the async clipboard API when there is a failure. The [`explainer.adoc`](https://github.com/w3c/clipboard-apis/blob/master/explainer.adoc) here doesn't use the rejection value at all, and the spec simply says "reject |p|".

https://github.com/w3c/clipboard-apis/blob/2ac3c0b5894558a4b468be6f09feb588efff4077/index.bs#L676

Is the intention that the Promise should simply reject with the value `undefined` instead of an `Error` instance?

This is how we have interpreted it here: feross/clipboard-copy#24

I feel that this goes against best practices, and I think there is a ton of code in the wild that does something similar to:

```js
runAllMythings().catch((err) => {
  console.log(err.stack)
})
```

Wouldn't it be nice to reject with an error, maybe `reject(new Error('Persmission for writing to clipboard was not granted'))`?

LinusU:
This was previously only done for the `read` method, this patch adds it to `readText`, `write` and `writeText` as well.

Fixes #76

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -639,7 +639,7 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 				
-				1. If |r| is not "granted", then reject |p|
+				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |data| be a copy of the [=system clipboard data=].
 				
@@ -673,7 +673,7 @@ urlPrefix: https://w3c.github.io/permissions/#dom-permissionstate-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard write permission=] [=in parallel=]
 				
-				1. If |r| is not "granted", then reject |p|
+				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |cleanItemList| be an empty {{DataTransferItemList}}.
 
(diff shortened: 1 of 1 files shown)


LinusU:
Hmm, not sure why Travis is failing, but it doesn't seem related to the changes:

> OSError: [Errno 2] No such file or directory: '/home/travis/build/w3c/clipboard-apis/bikeshed/bikeshed/spec-data/caniuse.json'

I have linked my GitHub account to my W3C account now, but I cannot retrigger the ipr check...

garykac:
I just re-ran the IPR check and is rejects with "LinusU did not make IPR commitments for this group." I'm not sure how you need to go about doing that.

LinusU:
Hmm, to be clear, I have no association with any company in the W3C group? Is that a prerequisite for making PRs to this repo? Could someone else make this change?

LinusU:
ping @garykac @gked, any chance of moving forward on this?

dontcallmedom:
Marked as non substantive for IPR from ash-nazg.

LinusU:
ping @garykac @gked, any chance of moving forward on this?

gked:
> ping @garykac @gked, any chance of moving forward on this?

sorry, for delay on this. let me try to reproduce the issue and see what is happening.

gked:
@LinusU hmm, travis built just fine when I integrated your changes in my own fork(see below).
@garykac if you don't mind, I'll accept the changes since I can attest to build passing.

image

PR #108, opened 2019-12-23
Use new HTML activation model

### LINKED ISSUE #107: Rephrase "direct user action" to match the new user activation model
We have User Activation v2 model in the HTML spec now. The current wording "running as a result of direct user action" should be changed accordingly. See the HTML issue mentioned below.

saschanaz:
Closes #107

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:

 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [x] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1597857)

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/clipboard-apis/pull/108.html" title="Last updated on May 4, 2020, 12:23 PM UTC (d82c76e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/108/e385af2...saschanaz:d82c76e.html" title="Last updated on May 4, 2020, 12:23 PM UTC (d82c76e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1017,8 +1017,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-read", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] does not
-			have [=transient activation=], false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] has
+		    [=transient activation=], false otherwise.
 
 		1. If |readWithoutGesture| is <a>granted</a>, then return true.
 
@@ -1053,8 +1053,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-write", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] does not
-			have [=transient activation=], false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=]
+		    has [=transient activation=], false otherwise.
 
 		1. If |writeWithoutGesture| is <a>granted</a>, then return true.
 
--- a/index.bs
+++ b/index.bs
@@ -1017,10 +1017,10 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-read", allowWithoutGesture: true }</code>
 			permission.
 
+		1. If |readWithoutGesture| is <a>granted</a>, then return true.
+
 		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] has
 		    [=transient activation=], false otherwise.
-
-		1. If |readWithoutGesture| is <a>granted</a>, then return true.
 
 		1. If |hasGesture| then,
 
--- a/index.bs
+++ b/index.bs
@@ -108,7 +108,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			With content such as mathematics, simply copying rendered text and pasting
 			it into another application generally leads to most of the semantics being
 			lost. MathML often needs to be transformed to be copied as plain text, for
-			example to make sure "to the power of" is shown with the caret  "^" sign
+			example to make sure "to the power of" is shown with the caret "^" sign
 			in a formula plain-text input. The XML source could also be placed in the
 			clipboard with the appropriate transformation occurring at paste time.
 			</li>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1017,8 +1017,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-read", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the current script is not running as a result of
```
saschanaz: Why does `hasGesture` become true when the script runs *without* user gesture, though?

marcoscaceres:
Should it consume the activation?

marcoscaceres:
@saschanaz, might be good to follow up here on the Mozilla side bug number and verification of what we do. Hopefully that will get things moving :)

saschanaz:
Seems Gecko already is using the new model, cc'ing @edgarchen and @smaug----

EdgarChen:
Yes, Gecko is now using the new model for clipboard, we just observe it, not consume. Blink seems to behave the same, https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/editing/commands/clipboard_commands.cc;l=77;drc=a29e33af63dd8814d4f08bf8e4882ae819144f35.

index.bs
marcoscaceres: I think that was a typo...

index.bs
```diff
@@ -1017,8 +1017,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
-			direct user action, false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] does not
+			have [=transient activation=], false otherwise.
```
marcoscaceres, line 1021:
```suggestion
		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] has
		    [=transient activation=], false otherwise.
```

index.bs
```diff
@@ -1053,8 +1053,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-write", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the current script is not running as a result of
-			direct user action, false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] does not
+			have [=transient activation=], false otherwise.
```
marcoscaceres, line 1057:
```suggestion
		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=]
		    has [=transient activation=], false otherwise.
```

index.bs
```diff
@@ -1017,8 +1017,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-read", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the current script is not running as a result of
-			direct user action, false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] has
+		    [=transient activation=], false otherwise.
 
 		1. If |readWithoutGesture| is <a>granted</a>, then return true.
```
saschanaz, line 1023: While at it:

```suggestion
		1. If |readWithoutGesture| is <a>granted</a>, then return true.

		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=] has
		    [=transient activation=], false otherwise.
```

marcoscaceres:
@saschanaz hopefully I didn't break it with my suggested changes there :/ ... last thing we are missing here is just confirming we have some WPT tests.

saschanaz:
The corresponding Gecko patch has no additional test success metadata, so probably no. Not sure how can this be tested though. There is a Chrome-only nonstandard API `navigator.userActivation` to query transient activation state, which surprisingly have several WPT tests.

marcoscaceres:
hehe, `navigator.userActivation` ... that's cute! ... but yeah, that's bad if that in WPT :(

index.bs
```diff
@@ -1053,8 +1053,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<code>{ name: "clipboard-write", allowWithoutGesture: true }</code>
 			permission.
 
-		1. Let |hasGesture| be true if the current script is not running as a result of
-			direct user action, false otherwise.
+		1. Let |hasGesture| be true if the [=relevant global object=] of [=this=]
+		    has [=transient activation=], false otherwise.
 
 		1. If |writeWithoutGesture| is <a>granted</a>, then return true.
```
dway123, line 1059: This should be before the previous step, to correspond with the previous instance of this in lines 1020-1024

dway123: For posterity, this was fixed in https://github.com/w3c/clipboard-apis/pull/122 (Thanks!)

PR #116, opened 2020-05-04
chore: remove index.html from master

saschanaz:
Per my understanding this file serves no purpose as currently the CI autogenerates it for `gh-pages` branch.


annevk:
You probably want to add `.gitignore` for `index.html` and maybe `out/` so it doesn't get added accidentally.

PR #117, opened 2020-05-04
Editorial: Export "allowed to modify the clipboard"

### LINKED ISSUE #115: Consider exporting "allowed to modify the clipboard" definition
Currently [execCommand spec uses the term](https://w3c.github.io/editing/ActiveDocuments/execCommand.html#the-copy-command) without explicit export from [this spec](https://w3c.github.io/clipboard-apis/#allowed-to-modify-the-clipboard). Since unexported terms are expected to only used within its parent spec, it should be explicitly exported to suppress the reference error from ReSpec.

saschanaz:
Fixes #115

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/clipboard-apis/pull/117.html" title="Last updated on May 4, 2020, 8:58 AM UTC (e800ae4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/117/090f0ae...saschanaz:e800ae4.html" title="Last updated on May 4, 2020, 8:58 AM UTC (e800ae4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -313,7 +313,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 	<h3 id="integration-with-other-scripts-and-events">Integration with other scripts and events</h3>
 
-		<h4 id="allow-modify-clipboard">Event handlers that are <dfn>allowed to modify the clipboard</dfn></h4>
+		<h4 id="allow-modify-clipboard">Event handlers that are <dfn data-export>allowed to modify the clipboard</dfn></h4>
 
 			Event handlers may write to the clipboard if any of the following is
 			true:


index.bs
```diff
@@ -313,7 +313,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 	<h3 id="integration-with-other-scripts-and-events">Integration with other scripts and events</h3>
 
-		<h4 id="allow-modify-clipboard">Event handlers that are <dfn>allowed to modify the clipboard</dfn></h4>
+		<h4 id="allow-modify-clipboard">Event handlers that are <dfn data-export>allowed to modify the clipboard</dfn></h4>
```
marcoscaceres, line 316: In BS, I think it's just "export" no? https://tabatkins.github.io/bikeshed/#dfn-export

```suggestion
		<h4 id="allow-modify-clipboard">Event handlers that are <dfn export>allowed to modify the clipboard</dfn></h4>
```

saschanaz: It's `data-export` [here](https://tabatkins.github.io/bikeshed/#dfn-contract). @tabatkins?

garykac: Sorry for not seeing this earlier. I already merged a fix for this with just "export", which I believe is correct here because this export isn't part of the data model.

saschanaz: I saw your commit, and `export` just is translated to `data-export` there. I think @tabatkins can clarify what's going on there and the documentation.

tabatkins: Yeah, in a bs source file you can just write `export`. It gets turned into `data-export` in the output, so that it's not an invalid attribute.

PR #120, opened 2020-06-13
Add Clipboard Feature Policy

### LINKED ISSUE #106: Restrict Clipboard API to top-level origin
The [Async Clipboard API](https://www.w3.org/TR/clipboard-apis/#async-clipboard-api) doesn't already restrict API use to top-level origins, but other potentially dangerous APIs like [screen share](https://w3c.github.io/mediacapture-screen-share/#feature-policy-integration-0) do. Using a feature policy to restrict usage to top-level origins should help avoid potential permission/data leakages across origins.

Could we please add a Feature Policy to require the Clipboard API to only be accessible to top-level frames, at least unless the owning origin explicitly allows subframes to access this? (An extension from this could be only allowing the top-level origin to access this API, and not allowing sub-frames to access this at all). This change can likely use very similar text as in the screen share spec.

dway123:
Closes #106 

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [x] Modified Web platform [tests](https://crrev.com/c/2164248)

Implementation commitment:

 * [x] Chromium (https://crrev.com/c/2164248)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10,7 +10,7 @@ TR: https://www.w3.org/TR/clipboard-apis/
 ED: https://w3c.github.io/clipboard-apis/
 Repository: w3c/clipboard-apis
 !Explainer: <a href="https://github.com/w3c/clipboard-apis/blob/master/explainer.adoc">Async Clipboard API Explainer</a>
-Editor: Gary Kacmarcik, Google, garykac@google.com, w3cid 59482
+Editor: Gary Kacmarcik, Google!!!, garykac@google.com, w3cid 59482
 Editor: Grisha Lyukshin, Microsoft, glyuk@microsoft.com, w3cid 78883
 Former Editor: Hallvord R. M. Steen, Mozilla
 Abstract:
--- a/index.bs
+++ b/index.bs
@@ -10,7 +10,7 @@ TR: https://www.w3.org/TR/clipboard-apis/
 ED: https://w3c.github.io/clipboard-apis/
 Repository: w3c/clipboard-apis
 !Explainer: <a href="https://github.com/w3c/clipboard-apis/blob/master/explainer.adoc">Async Clipboard API Explainer</a>
-Editor: Gary Kacmarcik, Google!!!, garykac@google.com, w3cid 59482
+Editor: Gary Kacmarcik, Google, garykac@google.com, w3cid 59482
 Editor: Grisha Lyukshin, Microsoft, glyuk@microsoft.com, w3cid 78883
 Former Editor: Hallvord R. M. Steen, Mozilla
 Abstract:
--- a/index.bs
+++ b/index.bs
@@ -630,7 +630,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. Let |f| be the result of checking clipboard feature policy [=in parallel=]
+
+				1. If |r| is not "granted", or |f| is not "granted", then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |data| be a copy of the [=system clipboard data=] represented as
 					a sequence of {{ClipboardItem}}s.
@@ -659,7 +661,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. Let |f| be the result of checking clipboard feature policy [=in parallel=]
+
+				1. If |r| is not "granted", or |f| is not "granted", then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |data| be a copy of the [=system clipboard data=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -630,7 +630,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. Let |f| be the result of checking clipboard feature policy [=in parallel=]
```
marcoscaceres, line 633: I think the feature policy checks can be done sync. This should link to [=allowed to use=] instead saying "result of checking".

marcoscaceres: Same for all the ones below...

marcoscaceres: I'd put the following as step 1 (link as appropriate):


> If the current settings object's responsible document is not allowed to use the "clipboard-read" feature, then return a promise rejected with "NotAllowedError" DOMException.

index.bs
```diff
@@ -1080,6 +1088,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 		</div><!-- algorithm -->
 
+<h2 id="feature-policy-integration">Feature Policy Integration</h2>
+
+This specification defines two [=policy-controlled feature=]s identified by the
```
marcoscaceres, line 1093:
```suggestion
This specification defines two [=policy-controlled feature|policy-controlled features=] identified by the
```

marcoscaceres:
again, it's kinda annoying that we ended up with "clipboard-read" and "clipboard-write" as the feature name :( ... would have been nice to have a single name, as we are again potentially setting a bad precedence.

index.bs
marcoscaceres: A minor thing... if you haven't already, try just [=policy-controlled features=] ... BS is pretty smart about matching plurals, so it might just work.

index.bs
dway123: Thanks! I think I've addressed this comment, but please let me know if it's not placed in the right spot not.

index.bs
dway123: Thanks! [=policy-controlled-features=] did work.

dway123:
Thanks for the review! I've updated the change a bit now per the comments.

Yeah, I agree that a single feature name would have been nice as well, but am just trying to match the clipboard spec's permission names with the feature policy, as feature policy and permissions are merging [since this change](https://github.com/w3c/webappsec-feature-policy/issues/359).

marcoscaceres:
[Filed bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1646270) on the Gecko side too.

index.bs
```diff
@@ -626,11 +626,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			1. Let |p| be a new [=Promise=].
 
+			1. If the [=current settings object's=] responsible document is not
+				[=allowed to use=] the "clipboard-write" feature, then reject |p|
+				with a "NotAllowedError" DOMException.
+
```
marcoscaceres, line 632: This current algorithm currently continues with the rejected promise, instead of returning it. Let's do: 

```suggestion
			1. If the [=current settings object's=] responsible document is not
				[=allowed to use=] the "clipboard-write" feature, then
				return a promise rejected with a "NotAllowedError" DOMException.

			1. Let |p| be a new [=Promise=].


```

index.bs
```diff
@@ -655,11 +659,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			1. Let |p| be a new [=Promise=].
 
+			1. If the [=current settings object's=] responsible document is not
```
marcoscaceres, line 662: Same as above... immediately return the rejected promise, otherwise continue in parallel.

index.bs
```diff
@@ -689,11 +697,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
```
marcoscaceres, line 700: As above here too...

index.bs
```diff
@@ -730,11 +742,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
```
marcoscaceres, line 745: Same... 😊

index.bs
marcoscaceres: Just some nits... I think you can link [=responsible document=] and [=a promise rejected with=] ... please confirm by looking up the terms at https://respec.org/xref/ ... you can probably link {{DOMException}} too.

saschanaz:
> again, it's kinda annoying that we ended up with "clipboard-read" and "clipboard-write" as the feature name :(

Are we too late to rename them? It seems currently Chrome has the only implementation, should we discuss about renaming?

marcoscaceres:
Yes, if only Firefox and Chrome are supporting these, we should discuss renaming them... but we would need to commit to implementing them too.

saschanaz:
Firefox has no implementation, only Chrome does.

marcoscaceres:
> Firefox has no implementation, only Chrome does.

Understood, but we should figure out if we will implement it in the future (and figure out if what's in the spec is palatable to us, and remove what Gecko is not willing to support to match reality).

wandyezj:
What is really annoying is that browsers have gone ahead with their own implementation for a critical feature. This ends up forcing developers to use commands to read and write to the clipboard to work around this nonsense.

PR #126, opened 2020-06-25
Change the ClipboardItem constructor to `ClipboardItemData`.

lgarron:
Right now:

- `ClipboardItemDataType` is a `string` or a `Blob`;
- `ClipboardItemData` is a *Promise* of a `ClipboardItemDataType`.

There are definitely some compatibility issues in the wild. In particular, Chrome only accepts `Blob`s, and throws an exception if you pass in a string or a Promise. Is it possible that the constructor is supposed to take `ClipboardItemDataType` values? If not, should it accept any values that are either `ClipboardItemDataType` *or* `ClipboardItemDataType`?

PR #127, opened 2020-06-25
New terminology: ClipboardItem entry as data type → value.

lgarron:
Right now, the spec refers to two things as "items":

- A list of `ClipboardItem`s.
- The individual data values inside a single `ClipboardItem`.

This is rather confusing, so this PR proposes to rename the latter to
"entries", where an entry conceptually corresponds to a data type (e.g.
`"text/string"`) that maps to a value (e.g. a Blob). This:

- Matches the JS naming convention (think of `Object.entries`).
- Does not change any spec semantics, only the names of some identifiers.

In addition, this PR also renames:

- `ClipboardItemDataType` → `ClipboardItemValue`, and
- `ClipboardItemData` → `ClipboardItemValuePromise`.

This matches the "entry is a type type mapping to a value" convention,
and makes it clear what is a Promise. The previous names may have caused
a semantic typo (https://github.com/w3c/clipboard-apis/pull/126).

Closes #????

 * [ ] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:


index.bs
```diff
@@ -589,16 +589,16 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
-		typedef Promise&lt;ClipboardItemDataType> ClipboardItemData;
+		typedef (DOMString or Blob) ClipboardItemValue;
+		typedef Promise&lt;ClipboardItemValue> ClipboardEntryValuePromise;
 
-		callback ClipboardItemDelayedCallback = ClipboardItemData ();
+		callback ClipboardItemDelayedCallback = ClipboardEntryValuePromise ();
 
 		[Exposed=Window] interface ClipboardItem {
-		  constructor(record&lt;DOMString, ClipboardItemData> items,
+		  constructor(record&lt;DOMString, ClipboardEntryValue> entries,
```
lgarron, line 598: This change assumes that https://github.com/w3c/clipboard-apis/pull/126 is correct (e.g. matching Chrome's current behaviour). If this should accept `Promise`s, then either:

- the record values should be `ClipboardEntryValuePromise` and Chrome's implementation should be changed, or
- the record values should be allowed to be either `ClipboardEntryValue` or `ClipboardEntryValuePromise`.

lgarron:
FYI: my W3C affiliation is set to GitHub, Inc., but this my participation in this repo should be considered under my Chromium affiliation. I don't know if there's a way to tell the CI this.

siusin:
@lgarron the IPR issue has been fixed by @travisleithead and our Systeam. It will no longer be a blocker for future contributions.

mbrodesser:
@lgarron: are you still working on this PR? The change itself is desirable, it just requires slight adaptation to match the current base.

mbrodesser:
@lgarron: just to be sure, do you still intend adapting this patch soon? It would be valuable.

mbrodesser:
@lgarron: can you please give an update if you still intend to adapt this patch?

travisleithead (approved):
These IDL changes don't impact any behavior, and they make sense to me.

Issue #75, opened 2018-03-22
Don't recommend user gesture in clipboard API


CendioOssman:
The new async clipboard API suggestion is a very nice step forward. However it still (vaguely) recommends user gestures as a way to prevent abuse. This method makes the clipboard API completely unusable for our use case so I'd like to plead for a recommendation against that in favour of other protection.

The use case here are remote desktop applications. The entire interface is server side so the Javascript is not aware of any of the logic, or even the layout. Hence it cannot connect user gestures with a clipboard update.

The flow would basically be:

1. The user presses Ctrl+C, resulting in a series of `keydown` events.
2. The page forwards these events to the remote system, exiting the event handler
3. The remote application triggers a Copy operation based on these events and sends some clipboard data to the client
4. The page receives the data and tries to update the local clipboard

This method is currently blocked, and would continue to be blocked unless browsers stop requiring a user gesture to access the API.

mbrodesser:
@CendioOssman: does it imply that all JS functionality requiring [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) is broken when using a remote desktop application like Cendio?

If there's are any news about this issue, please share them here.

CendioOssman:
Anything supporting transient activation should work fine in our scenarios, if I'm reading that description correctly¹. Assuming the transient activation duration is long enough to cover sending the notification over the network, having the remote applications react, and send a request back.

What won't work is things that must be done before returning from an event handler. We are simply too asynchronous for that. And it was my understanding that this was/is how some user activated things are blocked?

¹ There might be some corner cases, but I would assume 99% of applications only do clipboard operations as a result of a keyboard key or mouse button

mbrodesser:
Thanks for your fast response.

> Anything supporting transient activation should work fine in our scenarios, if I'm reading that description correctly¹. Assuming the transient activation duration is long enough to cover sending the notification over the network, having the remote applications react, and send a request back.
> 
> What won't work is things that must be done before returning from an event handler. We are simply too asynchronous for that. And it was my understanding that this was/is how some user activated things are blocked?
> 

Yes, Safari apparently requires calling `clipboard.readText()` from an event handler, which corresponds to the user gesture (see "Security and Privacy" at https://webkit.org/blog/10855/).

[...]

Issue #52, opened 2017-09-12
User gesture requirement for Clipboard API access


garykac:
We need to document how we want to handle the (current) User Gesture requirement once we have proper Permissions in place.

User gestures were put in place to help reduce drive-by API abuse since we didn't have a Permission mechanism in place. Now that we have Permissions, we could remove the gesture requirement, but some UAs may want to keep the gesture requirement so that they can auto-grant some access to the clipboard without requiring a Permission prompt.

If we want to include this explicitly in the spec, here is an initial proposal:
- Write: Gesture required for clipboard write UNLESS `fullAccess` Permission has already been granted
   - This is because the `fullAccess` permission is stronger than the basic permission
   - If basic `{ clipboard }` permission is DENIED, then write will not work (even with a gesture)
- Read: Since clipboard read is more dangerous, require `{ clipboard fullAccess=true }` Permission
   - Note: built-in (UA or OS-level) "paste" buttons may only require a user-gesture

garykac:
A first attempt at including gesture requirements and permissions is at:
https://w3c.github.io/clipboard-apis/#clipboard-permissions

gked:
@garykac in 9.1.1, what is the purpose of having to query the user for full access to the clipboard? Isn't the fact that systemPaste is set to true indicates that full access was meant to be granted?

garykac:
"full" is granting full read/write access to the clipboard without a gesture.
"read" and "write" access require a gesture.

This means that "full" is a bit more than simply "read" + "write"

garykac:
I've update the spec to include an explicit `requireGesture` param for the Permission and included an example of how a UA could use to this to provide a single "full" access permission.

travisleithead:
Related to our review: https://github.com/w3ctag/design-reviews/issues/222

travisleithead:
https://github.com/w3c/clipboard-apis/issues/52#issuecomment-383951308 this prevents web UI from initiating a copy/paste, which might be a deal-breaker...

garykac:
For implementations that want to gate this API on a user gesture, then that feels too restrictive to me (since, as you note, it will prevent custom UI for the clipboard operations).

Note that the existing clipboard event API already allows you to override the clipboard operation to the clipboard (see https://w3c.github.io/clipboard-apis/#clipboard-event-api). If people can only access the Write API inside the "copy" event (and Read from within Paste), then that's effectively the same restriction as the current API. 

[...]

PR #132, opened 2020-08-11
Remove clipboard-read permission

### LINKED ISSUE #101: Is "clipboard" permission replaced by "clipboard-read" and "clipboard-write"?
[The Permissions spec](https://w3c.github.io/permissions/#dom-permissionname-clipboard) still mentions `"clipboard"` but it seems it's removed from the Clipboard spec. Should the Permissons spec be updated to use the new permission names?

saschanaz:
Closes #101, see also https://github.com/w3c/permissions/issues/175

Implementation commitment:

 * [x] WebKit - Never shipped, and doesn't support this change.
 * [x] Gecko - Never shipped, and doesn't support this change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/pull/132.html" title="Last updated on Oct 26, 2020, 5:05 AM UTC (1084f2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/132/4231c62...1084f2e.html" title="Last updated on Oct 26, 2020, 5:05 AM UTC (1084f2e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -935,8 +935,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 	<a>powerful feature</a>s like the clipboard. It allows websites to request permissions
 	from users and query which permissions they have.
 
-	For the clipboard, two separate permissions are defined:
-	<dfn enum-value for="PermissionName">"clipboard-read"</dfn> and
+	For the clipboard, one permission is defined:
 	<dfn enum-value for="PermissionName">"clipboard-write"</dfn>
 
 	Note: Clipboard permissions currently only apply to the Async Clipboard API.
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
Thanks for doing this work @saschanaz.

pwnall:
Chrome has shipped the `clipboard-read` permission, so this would be a breaking change for us.

@dway123 do we have any data we could share about the magnitude of the breakage?

dway123:
Sure. The suggested removal of `clipboard-read` here implies [requiring user activation](https://github.com/w3c/clipboard-apis/issues/52) to gate Clipboard read. Therefore, this should break any site that currently doesn't require user activation to read the clipboard.

While we don't know the exact magnitude of the breakage (we don't currently have metrics on clipboard read use without user activation), we do know that:
- .03% of pageloads currently [use](https://chromestatus.com/metrics/feature/timeline/popularity/2369) `clipboard.read()`, and .22% of pageloads currently [use](https://chromestatus.com/metrics/feature/timeline/popularity/2371) `clipboard.readText()`. The subset of these sites that doesn't do so requiring user activation would be broken by this change.
- As of September 2020, 11,088 unique sites made use of the Async Clipboard API, according to the HTTP Archive. This number may be higher now, but also includes sites using clipboard write but not clipboard read, or clipboard read, but with user activation, so may represent more than the complete list of affected sites.
- Remote desktop applications, [including](https://github.com/w3c/clipboard-apis/issues/75) cendio.com, as well as Chrome Remote Desktop, would be broken by this change, as this would disallow them to propagate clipboard updates.

rniwa:
> Sure. The suggested removal of `clipboard-read` here implies [requiring user activation](https://github.com/w3c/clipboard-apis/issues/52) to gate Clipboard read. Therefore, this should break any site that currently doesn't require user activation to read the clipboard.

Aren't those sites also broken in Firefox & Safari? Or are you more concerned about Chrome only content?

dway123:
> Aren't those sites also broken in Firefox & Safari? Or are you more concerned about Chrome only content?

I suspect but have not verified these sites are also broken in Firefox and Safari, if they require user gesture for all Clipboard reads. I only discussed Chrome content because the question was regarding breakage in Chrome, and I only had public use information for Chrome, like [chromestatus feature use data](https://chromestatus.com/metrics/feature/timeline/popularity/2369). Please share information about breakage in Firefox & Safari, if possible, to provide more color for this conversation. Sorry, I'm not well acquainted with available public data for Firefox & Safari.

marcoscaceres:
> Please share information about breakage in Firefox & Safari, if possible, to provide more color for this conversation.

I don't think Firefox typically collects this kind of information (though some from Mozilla can correct me). They would need to check specific sites to see if they are breaking. However, likelihood is that they are not broken if the read is happening along with a user gesture... hopefully sites have figured that out already and at mitigating breakage by calling the API as a result of user activation.

saschanaz:
I can confirm that Mozilla doesn't have a use counter for `navigator.clipboard`.

marcoscaceres:
So, just chiming in a chair here. The appropriate course of action would be to merge this to match implementations. Then, as a followup, if folks from Google would still like to pursue this option, it might be worth sending a PR reverting the changes made in this PR.

rniwa:
> > Aren't those sites also broken in Firefox & Safari? Or are you more concerned about Chrome only content?
> 
> I suspect but have not verified these sites are also broken in Firefox and Safari, if they require user gesture for all Clipboard reads. I only discussed Chrome content because the question was regarding breakage in Chrome, and I only had public use information for Chrome, like [chromestatus feature use data](https://chromestatus.com/metrics/feature/timeline/popularity/2369). Please share information about breakage in Firefox & Safari, if possible, to provide more color for this conversation. Sorry, I'm not well acquainted with available public data for Firefox & Safari.

Safari does not collect such data either. Given Safari always requires user gesture to paste, it's pretty safe to assume such websites are broken in Safari. If those websites are broken in Safari, on the other hand, it means the website has some Chrome/Chromium specific behavior. Even in such a scenario, this will likely remain Chrome/Chromium only permissions so the two-implementation requirement means that this permission should not be added back to the spec.

marcoscaceres:
Ok, I'm going to go ahead and merge this, as the two engine requirement hasn't been met for this feature.

index.bs
```diff
@@ -960,87 +959,41 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 	There are 4 clipboard permissions:
```
yume-chan, line 960: It still says `4` instead of `2`

PR #143, opened 2021-05-20
Link to the HTML spec's "has focus steps".

mfalken:
This clarifies what "currently focused" means.

***
<a href="https://pr-preview.s3.amazonaws.com/mfalken/clipboard-apis/pull/143.html" title="Last updated on May 25, 2021, 1:16 AM UTC (a10caa5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/143/90ddec0...mfalken:a10caa5.html" title="Last updated on May 25, 2021, 1:16 AM UTC (a10caa5)">Diff</a>


mfalken:
@domenic @garykac can you review this? More context at https://github.com/jeremyroman/alternate-loading-modes/issues/55

index.bs
```diff
@@ -1171,7 +1171,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 		handlers) and data can be accessed in the absence of a user provided gesture.
 
 		To help prevent abuse, this API <em>must not</em> be available unless the script
-		is executing in the context of the currently focused document.
+		is executing in the context of a document that [=has focus steps|has focus=].
```
domenic, line 1174: Documents don't have focus. Specific elements inside them, or focusable areas, have focus.

Is this supposed to be about the browsing context having system focus?

domenic (changes requested):

domenic (approved):
LGTM, although ideally this would be in the actual algorithms instead of a magic patch at a distance from the privacy section of the spec. (That's a preexisting problem though, and not really in scope for you to fix...)

mfalken:
Thanks. I believe the original intent of the text is about system focus.  I'm not sure about one thing. Document.hasFocus() exists and uses https://html.spec.whatwg.org/multipage/interaction.html#has-focus-steps which looked convenient as it checks for system focus in the first step. I'm not sure if the other steps are needed. I just changed it to use system focus directly. Do you know what is going on here?

domenic:
The other steps handle iframe nesting cases. If you call `document.hasFocus()` inside an iframe, then this is true if:

- The iframe itself has focus within its parent (times N if it's multiply nested)
- The top-level BC has system focus

In non-iframe cases, target's BC = target's BC's TLBC, so candidate = target so the loop quits immediately.

mfalken:
I see. I wonder if the Clipboard spec also wants to disallow the API in iframes that don't have focus. In that case, would returning to "has focus steps" be better?

domenic:
Yes, that's a good point. Although you could make an argument for either behavior, probably the clipboard API spec authors intended the `document.hasFocus()` behavior.

I apologize for not reading your original PR closer. I think I assumed "has focus steps" was a link to https://html.spec.whatwg.org/#focused, but it's document-specific. So we should probably revert to your original.

mfalken:
No worries, I returned to "has focus steps" then. @garykac does this look good to you?

mfalken:
> No worries, I returned to "has focus steps" then. @garykac does this look good to you?

Sorry somehow I didn't realize @domenic has write access. I was assuming only the editors do. If you think it's OK I'm happy if you merge it. For the build error, I also have sent https://github.com/w3c/clipboard-apis/pull/147.

domenic:
Oh weird, I have write access, strange... I'd love it if we could get an editor to sign off before I touch anything though :).

garykac (approved):

garykac:
Adding links like this is great because it will make the (eventual) conversion to a more algorithmic spec easier. Thanks!

PR #147, opened 2021-05-23
Add link-default for HTML's Navigator interface.

### LINKED ISSUE #146: LINK ERROR: Multiple possible 'Navigator' interface refs.
This error happens on the PR at https://github.com/w3c/clipboard-apis/pull/143/checks?check_run_id=2643403146 but also seems to happen locally on the main branch:

```
$ bikeshed --die-on=link-error spec "index.bs" "index.bs.build.html" --md-status="WD" --md-date="2021-05-21"
LINK ERROR: Multiple possible 'Navigator' interface refs.
Arbitrarily chose https://html.spec.whatwg.org/multipage/system-state.html#navigator
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:interface; text:Navigator
spec:mediacapture-streams; type:interface; text:Navigator
<a data-lt="Navigator" data-link-type="interface">Navigator</a>
 ✘  Did not generate, due to fatal errors
```

The error does not occur if the --md-status and --md-date options are omitted.

mfalken:
Closes #146

***
<a href="https://pr-preview.s3.amazonaws.com/mfalken/clipboard-apis/pull/147.html" title="Last updated on May 23, 2021, 8:20 AM UTC (4515668)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/147/5b75462...mfalken:4515668.html" title="Last updated on May 23, 2021, 8:20 AM UTC (4515668)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -32,6 +32,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/interaction.html#; type: dfn;
 	text: dom anchor
 url: https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2; type: dfn;
 	text: the body element
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/system-state.html; type: interface;
+	text: Navigator
 urlPrefix: http://www.w3.org/TR/html5/browsers.html#; type: dfn;
 	text: allowed to show a popup
 urlPrefix: https://dom.spec.whatwg.org/#; type: dfn;


index.bs
```diff
@@ -32,6 +32,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/interaction.html#; type: dfn;
 	text: dom anchor
 url: https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2; type: dfn;
 	text: the body element
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/system-state.html; type: interface;
```
marcoscaceres, line 35: Silly question for those that know BS, but why does BS not auto link these? Same with the next one that points to "w3c/html5" or the DOM spec? My understanding is that we shouldn't need any of these hard links.

mfalken: It looks like Bikeshed can auto-link it but according to the error in #146 it needs this because it can't decide between linking to HTML's or Media Capture Streams's definition of `Navigator`.

But I don't understand why, because Media Capture Streams seems to only declare a `partial interface Navigator`:
https://w3c.github.io/mediacapture-main/#navigator-interface-extensions

Also I didn't get the error when removing --md-status="WD" --md-date="2021-05-21".

marcoscaceres: @tabatkins, could you provide us with some guidance by chance? It would be great if we could get rid of the entire `<pre class=anchors>`  (with the exception of ecma262, probably).

garykac (approved):

index.bs
tabatkins: Well, you don't need an `anchors` block to handle this in any case; when the link exists, and you just want to disambiguate them by default, a `link-defaults` block does so much more safely and easily.

```
<pre class=link-defaults>
spec: html; type: interface; text: Navigator
</pre>
```

Bikeshed will suggest these lines to you when there is such a conflict:

```
LINK ERROR: Multiple possible 'Navigator' idl refs.
Arbitrarily chose https://html.spec.whatwg.org/multipage/system-state.html#navigator
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:interface; text:Navigator
spec:mediacapture-streams; type:interface; text:Navigator
{{Navigator}}
```

(Note that you'll only see this conflict when building a "snapshot" spec, like a WD, rather than a "current" spec, like an ED. Since Media Capture's ED is ReSpec-on-the-client, Bikeshed can't find most/all of its definitions, so the only existing "current" definition, from HTML, wins for "current"-type specs. Only when you build for "snapshot" do we see the two definitions on equivalent footing and get the conflict.)

I suggest deleting the entire anchors block and seeing what Bikeshed actually complains about, then just fixing those terms, using a link-defaults when Bikeshed suggests such. You likely will be left with only JS terms; if any W3C specs come up blank, verify that they're in <https://github.com/tabatkins/bikeshed-data/blob/main/data/specs.json> and if not, file a bug on me to add them.

-----

> But I don't understand why, because Media Capture Streams seems to only declare a partial interface Navigator:

Unfortunately, Media Capture *does* have a definition for the Navigator interface itself; not in the IDL, but in the *heading for that section*. That's a bug in the spec's markup.

marcoscaceres: CC @anssiko ^^^

sidvishnoi: > Since Media Capture's ED is ReSpec-on-the-client, Bikeshed can't find most/all of its definitions

@tabatkins would it be possible to use spec-generator instead of the raw ReSpec spec to parse the definitions?

PR #155, opened 2021-09-02
Add mandatory and optional data types list.

snianu:
Closes https://github.com/w3c/editing/issues/305

Implementation commitment:

 * Already implemented by WebKit, Chromium & Gecko.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -528,19 +528,17 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			corresponding native type exists on the clipboard:
 
 			* text/plain
-			* text/uri-list
-			* text/csv
-			* text/css
 			* text/html
-			* application/xhtml+xml
 			* image/png
+
+			These data types may be exposed by <em>paste</em> events if a
+			corresponding native type exists on the clipboard:
+
+			* text/uri-list
 			* image/jpg, image/jpeg
 			* image/gif
 			* image/svg+xml
-			* application/xml, text/xml
-			* application/javascript
-			* application/json
-			* application/octet-stream
+			* text/rtf
 
 		<h4 id="writing-to-clipboard">Writing to the clipboard</h4>
 
@@ -549,12 +547,18 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<em>copy</em> and <em>cut</em> events.
 
 			* text/plain
-			* text/uri-list
-			* text/csv
 			* text/html
+			* image/png
+
+			These data types may be placed on the clipboard with a corresponding
+			native type description if added to a {{DataTransfer}} object during
+			<em>copy</em> and <em>cut</em> events.
+
+			* text/uri-list
+			* image/jpg, image/jpeg
+			* image/gif
 			* image/svg+xml
-			* application/xml, text/xml
-			* application/json
+			* text/rtf
 
 			Advisement: Warning! The data types that untrusted scripts are allowed to write to the
 			clipboard are limited as a security precaution. Untrusted scripts can


annevk:
I would prefer to only land the MUST lists as those seem to have agreement. Optional support for types doesn't seem like a great thing to allow and wasn't discussed in the referenced issue.

snianu:
@annevk 
Chromium, Safari & FF have support for types that are not in the list, but they are exposed via DataTransfer & async clipboard APIs. 
e.g. Safari has support for `text/uri-list`, Chrome has support for `image/svg+xml` etc.
I think it's fine to add the mime types that are supported in all browsers in the MUST support list, but we should at least mention the types that are supported by a particular browser so the web authors are aware of it. This type may not be available in other browsers so I added those "optional" types to the MAY support list. It also leaves room for future improvements to support more standard well known types defined by the OS (this is different than custom formats support).

annevk:
That's not entirely clear to me. For the types that only some browser supports the right answer might well be that the browser removes support for that type. If all browsers are to implement it, it shouldn't be optional. This isn't a guide for web developers, this is what we want browsers to converge on.

snianu:
@annevk OK, removed the optional types. PTAL.

annevk (commented):
This looks to be in line with the discussion, however, shouldn't JPEG and GIF be supported as well?

This would benefit from at least one other set of eyes.

snianu:
Added @mkruisselbrink to the review as well.

snianu:
> shouldn't JPEG and GIF be supported as well

I think in Chromium, we convert these into PNG and write it to the clipboard. @a-sully made changes in this space so maybe he can confirm the behavior here.

a-sully:
I'm not familiar with the details of how non-PNG images are written to the clipboard, but I can confirm that PNG is the only image format supported on the Chromium clipboard. That being said, JPEGs and GIFs are typically copied to and read from the clipboard as files using the `text/uri-list` mime type anyways.

snianu:
@a-sully 
> That being said, JPEGs and GIFs are typically copied to and read from the clipboard as files using the text/uri-list mime type anyways

At least on Windows, I don't see Chromium reading GIF, JPEGs. In [this ](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_win.cc;drc=8aa9dc97551d28dec43afebf525d83bc2aca1c3d;l=481)method, I see only PNG and BMP formats being read from the clipboard. Do we read these formats somewhere else? If not, how do we read those images in `text/uri-list`?

a-sully:
They're read as files using the [`ReadFilenames()` code path](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_win.cc;l=530;drc=8aa9dc97551d28dec43afebf525d83bc2aca1c3d). Though it seems that Windows ([`CFSTR_FILENAMEW`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_format_type_win.cc;l=261;drc=6997f8a177359bb99598988ed5e900841984d242;bpv=1;bpt=1)) and Mac ([`NSFilenamesPboardType`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_format_type_mac.mm;l=97;bpv=1;bpt=0)) actually use custom types rather than `text/uri-list` (which is used on Linux + Android), so I take that back. Should've confirmed in the code first.

I realized I wasn't very clear on this above. Yes, you are correct. Reading files is not yet supported for the Async Clipboard API, though we'd like to at some point (see the [tracking bug](https://crbug.com/1099775)).

snianu:
Thank you for confirming! Since Safari already supports `text/uri-list` MIME type and Chrome has plans to add this support in the future, should we add this format to the mandatory data types?

snianu:
Gentle ping.. I'd like to merge this PR and work on improving other parts of the spec.

whsieh:
Looks good to me!

PR #157, opened 2021-09-16
Update README.md

xfq:
Update README.md because the group is now using GitHub Actions to build/validate/publish the spec.


xfq:
(The CI is red though. https://github.com/tabatkins/bikeshed/pull/2139 might be related.)

snianu (approved):

PR #158, opened 2021-10-05
Add clipboard IDL description.

snianu:
Closes #135

For normative changes, the following tasks have been completed:

 * [ ] Modified Web platform tests (link to pull request)

Implementation commitment:

 * [X ] WebKit Shipped
 * [X ] Chromium Shipped
 * [X] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1619947)

Issue #150, opened 2021-06-28
Make async clipboard APIs (read/write) to sanitize interoperably with setData/getData for text/html


gked:
Hi All,
Excel online team reached out to us on several issues they found with clipboard formats. Sanitization across legacy and async clipboard APIs is not consistent for some mime types. For example, while text/html when set with setData/getData, keeps meta tags in Chrome, async clipboard APIs strip them down. 
This creates an issue for online text editing applications. Target apps, processing paste, rely on meta tags to infer information about the clipboard payload source. @snianu  has wrote a detailed analysis which can be found [here](https://docs.google.com/document/d/1rTEg2I-hMPXGiLrEMqKJz2Ycu6GRjlM3uvakOe84m8Q/edit#heading=h.yi2bvmwbg260). 
A matrix view of formats can be found [here](https://docs.google.com/spreadsheets/d/16GxpRzx77NaLTKsU5e1CZVtA5ehvxpZVFmJ-02A0uxs/edit?usp=sharing).
We propose that we make async clipboard read/write serialization behaviors consistent with legacy clipboard API such as setData/getData.
The reason being browsers already expose this information through setData/getData and making async clipboard read/write to behave in a compatible way which will ease its adoption over time. 

If we can all agree on this, we will follow up with a PR clarifying the behavior in the Clipboard Spec which will also take care of #140 

CC: @rniwa @whsieh  @dway123 @garykac @BoCupp-Microsoft @megangardner @johanneswilm

dway123:
I'm supportive of having `datatransfer` and `navigator.clipboard` (async clipboard) APIs be consistent in their sanitization behavior of text/html (and other formats). Thank you for the detailed analysis, and for offering to help update the spec to clarify this!

annevk:
How did the discrepancy arise in Chrome? Part of the plan with https://wicg.github.io/sanitizer-api/ is to get this all standardized. Does that meet the requirements you have?

cc @mozfreddyb @otherdaniel

otherdaniel:
> Part of the plan with https://wicg.github.io/sanitizer-api/ is to get this all standardized. Does that meet the requirements you have?

I don't know much about the specific question here, but on clipboard vs Sanitizer API: This is a v2 item. One major block of functionality missing is that the clipboard sanitizers do some form of style normalization, while the Sanitizer is presently HTML only.

[...]

PR #162, opened 2021-11-04
Add Custom Clipboard format support to async clipboard.

snianu:

 * [X] Modified Web platform tests (https://wpt.fyi/results/clipboard-apis?label=master&label=experimental&aligned&q=async%20clipboard)

Implementation commitment:

 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=106449)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/pull/162.html" title="Last updated on Nov 4, 2021, 4:37 PM UTC (1cda092)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/162/0f67e6a...1cda092.html" title="Last updated on Nov 4, 2021, 4:37 PM UTC (1cda092)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -168,7 +168,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 		exposed as a {{DataTransfer}} object that mirrors the contents of the clipboard.
 
 	* For the [[#async-clipboard-api]], the [=system clipboard data=] is
-		exposed as a sequence of {{ClipboardItem}} objects that mirrors the contents of
+		exposed as a sequence of [=ClipboardItem=] objects that mirrors the contents of
 		the clipboard.
 
 <h2 id="clipboard-events-and-interfaces">Clipboard Events</h2>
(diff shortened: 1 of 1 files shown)


tomayac:
At the risk of sounding incredibly stupid but with the excuse of being an _English as a second language_ person: what is the background of the word [pickling](https://www.google.com/search?q=pickling&tbm=isch&ved=2ahUKEwiyk-i74P7zAhUDG-wKHaKIBWMQ2-cCegQIABAA&oq=pickling&gs_lcp=CgNpbWcQAzIFCAAQgAQyBQgAEIAEMgUIABCABDIFCAAQgAQyBQgAEIAEMgUIABCABDIFCAAQgAQyBQgAEIAEMgUIABCABDIFCAAQgAQ6BwgjEO8DECc6BAgAEENQAFitCGDSCmgAcAB4AIABzgGIAawJkgEFMC43LjGYAQCgAQGqAQtnd3Mtd2l6LWltZ8ABAQ&sclient=img&ei=DdiDYfLYNoO2sAeikZaYBg&bih=859&biw=1680) 🥒? Or is it about the [tool](https://en.wikipedia.org/wiki/Pickaxe) ⛏ to pickle the strong edges off of dangerous clipboard items?

snianu:
@tomayac Here is some additional background :) https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#additional-background

I think I should just change the name to clipboard custom formats as it could be confusing to many. Thanks again for your feedback!

tomayac:
> @tomayac Here is some additional background :) https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#additional-background

#TIL! Oh my God… This is super geeky, but completely inaccessible to the uninitiated. No I know.

> I think I should just change the name to clipboard custom formats as it could be confusing to many. Thanks again for your feedback!

Definitely consider renaming. Danke!

annevk:
This depends on #158, right? And we also still have a discussion going on over in #150 as to whether these sanitization options make sense. It's not clear to me how this is ready for evaluation (or shipping).

zbyte-king (approved):

zbyte-king (approved):

snianu:
@annevk 
> This depends on #158, right?

Yes, I think the spec is now in a good state after the recent changes. Will address Domenic's recent comments and then this will be ready for review.

> And we also still have a discussion going on over in #150 as to whether these sanitization options make sense. It's not clear to me how this is ready for evaluation (or shipping).

#150  is related to writing unsanitized HTML content under the standard HTML format, not the custom format that this feature is about. I think we should evaluate this separately as custom formats are unsanitized content by definition. Also we had this discussion in multiple meetings with Apple and Google folks, and we all agreed to add this feature to the spec in a non-normative note. Here are the meeting minutes: https://github.com/w3c/editing/issues/334#issuecomment-933939592

rniwa:
> > And we also still have a discussion going on over in #150 as to whether these sanitization options make sense. It's not clear to me how this is ready for evaluation (or shipping).
> 
> #150 is related to writing unsanitized HTML content under the standard HTML format, not the custom format that this feature is about. I think we should evaluate this separately as custom formats are unsanitized content by definition. Also we had this discussion in multiple meetings with Apple and Google folks, and we all agreed to add this feature to the spec in a non-normative note. Here are the meeting minutes: [w3c/editing#334 (comment)](https://github.com/w3c/editing/issues/334#issuecomment-933939592)

I don't see how? This PR adds `unsanitized` as a dictionary option to `ClipboardItem`. I see no reason that's necessarily if we don't add `unsanitized` in the first place.

snianu:
> I don't see how? This PR adds unsanitized as a dictionary option to ClipboardItem. I see no reason that's necessarily if we don't add unsanitized in the first place.

We added `unsanitized` to `ClipboardItemOptions` as it is optional and we don't want to create yet another member to track the unsanitized list. We don't need the `unsanitized` list during `write()`, we only need it for `read()`. I guess we could also make the `unsanitized` member optional in the dictionary. That way we could reuse this type without breaking anything? Let me know what you think.. These will be added as a non-normative note to the spec so browsers may or may not support this type.

rniwa:
> > I don't see how? This PR adds unsanitized as a dictionary option to ClipboardItem. I see no reason that's necessarily if we don't add unsanitized in the first place.
> 
> We added `unsanitized` to `ClipboardItemOptions` as it is optional and we don't want to create yet another member to track the unsanitized list. We don't need the `unsanitized` list during `write()`, we only need it for `read()`. I guess we could also make the `unsanitized` member optional in the dictionary. That way we could reuse this type without breaking anything? Let me know what you think.. These will be added as a non-normative note to the spec so browsers may or may not support this type.

I don't see how that makes sense for a Web API. It means if someone writes code using Safari, it may not work in Chrome / Edge. Generally, we'd like to avoid that sort of thing if possible. My perspective on this PR is that we need to resolve [the issue ](https://github.com/w3c/clipboard-apis/issues/150) first.

snianu:
> I don't see how that makes sense for a Web API. 

We have several concepts in the `ClipboardItem` object that don't make sense on other platforms except iOS/iPadOS/MacOS. e.g.
1. We have a `presentationStyle` member in `ClipboardItemOptions` that is only supported in iOS/iPadOS. Chromium doesn't support this attribute and has no plans to support this type in the future.
2. We have multiple `ClipboardItem` support that doesn't make sense on platforms like Windows, Android, ChromeOS etc. 

Yet we chose to add it to the async clipboard API anyways, so I think we can add the `unsanitized` option as an optional type in `ClipboardItemOptions` as we already have a precedent to add options that may never be implemented/supported on other platforms.

rniwa:
> Yet we chose to add it to the async clipboard API anyways, so I think we can add the `unsanitized` option as an optional type in `ClipboardItemOptions` as we already have a precedent to add options that may never be implemented/supported on other platforms.

Okay, we can agree to disagree on this. We'd submit an objection when the time comes.

zbyte-king:
I thought it was weird myself but im just starting all this and don't know
a lot im glad you both think so as well

slightlyoff:
It's pretty unhelpful for us to be debating the shape of the API without example code and an explainer; it makes it somewhat difficult to judge how the proposed changes would be used.

@snianu: are you able to write up an explainer using the TAG's template? Example code in particular would be helpful:

https://github.com/w3ctag/tag.w3.org/blob/main/explainers/index.md

snianu:
@slightlyoff 
> @snianu: are you able to write up an explainer using the TAG's template? 

Sorry, forgot to mention in the PR. Here is the explainer: https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md

mbrodesser:
> @annevk
> 
> > This depends on #158, right?
> 
> Yes, I think the spec is now in a good state after the recent changes. Will address Domenic's recent comments and then this will be ready for review.
> 
> > And we also still have a discussion going on over in #150 as to whether these sanitization options make sense. It's not clear to me how this is ready for evaluation (or shipping).

Let's please finish #158 first and not rush these changes in.

zbyte-king (approved):

snianu:
Since we cannot come to an agreement, we have decided to abandon this PR. We will write about the new custom format map and the `unsanitized` option in a separate document which wouldn't be part of the clipboard API spec.

zbyte-king:
https://bugs.chromium.org/p/chromium/issues/detail?id=150835#c73

annevk:
Hey @snianu, this seems really weird. Overall this is a feature that has buy-in from everyone, but there's a small aspect that doesn't. Is the idea now to not standardize it and just ship something in Chromium? Doesn't that kind of defeat the point of having a standards venue?

(I'm not sure why I deleted the branch, that was not intentional.)

snianu:
@annevk I think the main point of contention was the `unsanitized` option which we would like to keep since we want to support sharing of the custom format content across origins(e.g. from Excel online to PPT to Word online to Outlook etc) and able to write unsanitized HTML using the standard HTML format. For Safari it is moot because they don't want to support that scenario. In [our meetings ](https://github.com/w3c/editing/issues/334#issuecomment-933939592) with Apple folks, they raised the concern that a web spec shouldn't define platform specific clipboard format naming convention/format structure -- They wanted the OS to specify the custom formats. We think that these custom formats are special as they are generated by the web browsers, and the OS (have only checked Windows, CrOS, Android, Linux and Mac) already has support for custom formats, so standardizing at least the [custom format map](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#os-interaction-format-naming) shouldn't be an issue. I wouldn't say that they were completely opposed to the custom format concept in general (and neither are you :)) but if we don't want to standardize the [custom format map](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#os-interaction-format-naming) and `unsanitized` option, then I'm not sure what changes need to be made in the spec?

annevk:
We don't have to define the exact format the OS clipboard uses, but we can say that these entries are special in that native apps don't recognize them. So it appears as `application/example` to websites, but not native apps. Defining for which formats that happens seems important for interoperability.

I do wonder how @rniwa et al envision the macOS clipboard to work in a way that native apps can work with Chrome/Firefox clipboard entries just as well as Safari entries. But if the macOS clipboard is extended for these kind of web entries it seems Chrome/Firefox could support that as well, in theory.

snianu:
@annevk 
> Defining for which formats that happens seems important for interoperability.

Agreed. We do want to define the [custom format map](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#os-interaction-format-naming) naming and its structure in the spec for interop reasons. Is it possible to discuss in the next EditingWG meeting? Would you be able to attend? Apple made it clear that they were not interested in implementing this new custom format feature, but with at least Chromium and Firefox support, I think we can add it to the official clipboard API spec using the MAY language instead of MUST?

annevk:
I've added https://www.w3.org/events/meetings/a97eea9a-0249-4da8-888a-95edc68aa65f/20220114T090000 to my calendar. Let me know if we cannot discuss it as it's clashing a bit with my normal evening.

snianu:
@annevk thank you! I've added the agenda label but tagging few folks to this issue as FYI. @BoCupp-Microsoft @travisleithead @johanneswilm @whsieh

annevk:
I've put forward an alternative proposal in #165.

pwnall:
I think this PR had some non-controversial parts that seem valuable. Would it make sense to try to get those merged separately?

For example, I think that the added lines 610-616, 623-648 (minus 636) document Async Clipboard API behavior that is either shipped or non-controversial (not related to sanitization, custom types, or permissions).

I'm hoping this would make it easier to compare proposals for the controversial pieces.

snianu:
@pwnall Yes, I created a separate PR to merge those changes into the clipboard API spec: https://github.com/w3c/clipboard-apis/pull/158

PR #164, opened 2022-01-16
Remove clipboard-write permission

### LINKED ISSUE #163: Should clipboard-write permission be removed too?
https://github.com/w3c/permissions/issues/175#issuecomment-671553031 from @rniwa:

>Shouldn't we also remove "clipboard-write" given we wouldn't have multiple implementation support for that either??

And I think Mozilla also have no intention to implement that. (cc @mbrodesser, @annevk)

saschanaz:
Closes #163 

Implementation commitment:

 * [x] WebKit https://github.com/w3c/permissions/issues/175#issuecomment-671553031
 * [x] Gecko https://github.com/w3c/permissions/pull/246#issuecomment-863022451

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -618,11 +618,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. If |r| is not true, then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |data| be a copy of the [=system clipboard data=] represented as
 					a sequence of {{ClipboardItem}}s.
-					
+
 					Note: As further described in [[#image-transcode]] this explicitly does not transcode images.
 					Rather the original unmodified image data should be exposed to the website.
 
@@ -650,7 +650,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. If |r| is not true, then reject |p| with a "NotAllowedError" DOMException
 
 				1. Let |data| be a copy of the [=system clipboard data=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -618,11 +618,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard read permission=] [=in parallel=]
```
saschanaz, line 619: Should this still be in parallel? I guess not? Out of scope here anyway.

marcoscaceres: agree... it's already "in parallel"..

```suggestion
				1. Let |r| be the result of running [=check clipboard read permission=]
```

index.bs
```diff
@@ -618,11 +618,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
 
-				1. If |r| is not "granted", then reject |p| with a "NotAllowedError" DOMException
+				1. If |r| is not true, then reject |p| with a "NotAllowedError" DOMException
```
marcoscaceres, line 621: does Bikeshed support?:

```suggestion
				1. If |r| is not true, then reject |p| with a "NotAllowedError" {{DOMException}}
```

index.bs
```diff
@@ -993,34 +946,17 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
-				<code>{ name: "clipboard-write", allowWithoutGesture: false }</code>
-				permission.
-
-				Note: User agents may choose to request a stronger permission that
-				will implicitly update this permission.
-
-		1. Return the result of [=request permission to use=] the
-			<code>{ name: "clipboard-write", allowWithoutGesture: true }</code>
-			permission.
+		1. Return false.
```
marcoscaceres, line 959: Maybe just for clarity:

```suggestion
		1. Otherwise, return false.
```

marcoscaceres (approved):
Just a couple of nits. Otherwise looks good to me.

snianu:
Curious, why are we removing the clipboard permissions from the spec? Chromium browsers support this and we have no plans on removing it. In fact we will be extending this to pickling APIs so I don't see why this should be removed from the spec? There are concepts in the spec that is only applicable to Apple platforms (presentation style, multiple clipboard items etc) and we have no plans on supporting those in the future, and yet we still added to the spec so I don't see any rationale as to why we are removing something that is currently supported by two browsers (Edge and Chrome). Consider this a formal objection from Microsoft. @mkruisselbrink @a-sully to chime in from Google.

annevk:
It's being removed because it was removed from the Permissions specification, per a previous discussion.

Also, Chrome and Edge do not count as two implementers per any usual definition of that word. That seems rather disingenuous.

As for Apple platform specific features, are those not different in that multiple browsers could support them, but they would only have an effect on Apple platforms? If there's no support from multiple implementers though perhaps those features need to be reconsidered indeed. Best discussed in its own issue.

snianu:
> Also, Chrome and Edge do not count as two implementers per any usual definition of that word

I'm not sure where you pulled that definition from, but we have complete control over what gets shipped in Edge regardless of the Blink API shipping process. As an example, we have shipped Pickling API in Edge and we are still waiting on the approval from the Blink API owners to ship in Chromium (which may or may not be approved), but considering that there is a strong developer support from Office products, Adobe, Figma, Google products etc, I'm hoping that it will get approved.

Removing this from the spec makes us non-compliant, and considering the fact that we are the ones maintaining this spec(https://github.com/w3c/clipboard-apis/pull/158), I don't think we would be in favor of this change. If this is the direction Mozilla and Apple want to take, then you are right when you said that "It's a failure of the standards committee".

annevk:
What typically counts as an implementation for the purposes of standards is a fully independent implementation. A single implementation that two products ship parts of with different shipping priorities is just not it, I'm sorry.

snianu:
> What typically counts as an implementation for the purposes of standards is a fully independent implementation

Can you please be a little more specific? What exactly is the formal definition of a "fully independent implementation"? Blink is a fork from Webkit so it's not exactly independent(according to what you trying to imply in your arguments).

annevk:
Right and for parts where they would use the same code to implement something that would not count.

snianu:
well, I'm not gonna get into the philosophical arguments about what "same code" even means, but I'm adding the agenda label to this PR so we can discuss with the Editing WG before making any changes to the spec that is in the Editing WG charter.

marcoscaceres:
As a neutral observer here, a different way of framing this is that two implementers are objecting to its inclusion. For that reason this probably should not have landed in the original spec. It may be worth investigating how this landed in the spec in the first place, and wether sufficient review/feedback was gathered in that original pull request (in alignment with the W3C Process, and with respect to consensus building).

artem1982:
It was marked as Pri1 bug on chromium: 
https://bugs.chromium.org/p/chromium/issues/detail?id=1334203

garykac:
FWIW, the current spec fails to build because bikeshed can no longer find a definition for 'permission task source'.

We need to either:
(a) Keep the permission steps in the algorithm, but update it to remove the reference to this term
(b) Remove all the permission steps
(c) <something else>?

Doing (a) seems problematic since it wouldn't really be a complete algorithmic description.

My preference would be for the spec to be updated to allow, but not require, an implementation to gate access to this feature based on a permission or user gesture or whatever.

Is that an acceptable middle ground?

marcoscaceres:
this is related to https://github.com/w3c/permissions/pull/407 ... I think there is some guidance there.

PR #171, opened 2022-02-24
[meta] Fix broken markup

tabatkins:
Your `dl` was unclosed. Went ahead and switched it to Markdown, as that requires less tracking anyway.


snianu (approved):
Thanks for working on this!

PR #174, opened 2022-05-04
Remove sanitize copy requirement from the spec.

### LINKED ISSUE #150: Make async clipboard APIs (read/write) to sanitize interoperably with setData/getData for text/html

snianu:
Closes #150 

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/pull/174.html" title="Last updated on May 20, 2022, 6:50 PM UTC (467a4e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/174/14c72dc...467a4e6.html" title="Last updated on May 20, 2022, 6:50 PM UTC (467a4e6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -786,14 +786,14 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 					1. Abort these steps.
 
-				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the sanitized copy of text/html format, but image/png format has unsanitized payload to preserve meta data.
+				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the content of text/html format, but image/png format has unsanitized payload to preserve meta data.
 
 					Issue: Some OSs contain multiple clipboard (e.g. Linux, "primary", "secondary", "selection"). Define from which of those data is read.
 
+					Issue: Currently Safari does some sanitization for [=mandatory data types=]. See https://github.com/w3c/clipboard-apis/issues/150.
+
 					Note: As further described in [[#image-transcode]] this explicitly does not transcode images.
 					Rather the original unmodified image data should be exposed to the website.
-
-					Issue: Add definition of sanitized copy.
 
 				1. [=Queue a global task=] on the [=clipboard task source=], given |realm|'s [=Realm/global object=], to perform the below steps:
 
@@ -837,11 +837,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 					1. Abort these steps.
 
-				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the sanitized copy of text/plain format.
+				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the content of text/plain format.
 
 					Issue: Some OSs contain multiple clipboard (e.g. Linux, "primary", "secondary", "selection"). Define from which of those data is read.
 
-					Issue: Add definition of sanitized copy.
+					Issue: Currently Safari does some sanitization. See https://github.com/w3c/clipboard-apis/issues/150.
 
 				1. [=Queue a global task=] on the [=clipboard task source=], given |realm|'s [=Realm/global object=], to perform the below steps:
 
@@ -953,17 +953,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 							1. If |type| is not in the [=mandatory data types=] list, then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm| and abort these steps.
 
-							1. Let |cleanItem| be a sanitized copy of |blob|.
+							1. Append |blob| to |cleanItemList|.
 
-								Issue: Add definition of sanitized copy.
-
-							1. If unable to create a sanitized copy, then follow the below steps:
-							
-								1. [=Reject=] |p| with {{"NotAllowedError"}} DOMException in |realm|.
-
-								1. Abort these steps.
-
-							1. Append |cleanItem| to |cleanItemList|.
+							Issue: Currently Safari does some sanitization. See https://github.com/w3c/clipboard-apis/issues/150.
 
 						1. Let |option| be |clipboardItem|'s [=ClipboardItem/clipboard item=]'s [=presentation style=].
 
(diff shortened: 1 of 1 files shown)


snianu:
@mbrodesser @annevk @BoCupp-Microsoft FYI. I added a note for Safari's sanitization behavior, but since FF and Chromium are interop and we have a desire to align async api with the DataTransfer API, I'm removing the sanitization language from the spec only for async apis. PTAL.

annevk:
I don't think we should diverge on the internal details between these APIs. I don't really see what benefit that brings.

snianu:
@annevk This change aligns async clipboard HTML read/write behavior with DataTransfer APIs for Chromium and Firefox browsers. Sanitization requirement in async clipboard APIs is a regression for us since popular apps like Office Online, Google docs, Adobe etc rely on clipboard APIs to not strip out content that it thinks is "harmful". Since Safari is the only browser that has a different behavior, any objection to making this change in the spec?

annevk:
Yeah, I think this goes against what we discussed in the meetings.

snianu:
In [this ](https://github.com/w3c/clipboard-apis/issues/150#issuecomment-1031684598)comment, you mentioned that you are in favor of aligning the behavior of DataTransfer and Async clipboard APIs. Am I misunderstanding something?

annevk:
Yes and @whsieh and I argued for allowing sanitization at least for built-in types too. Those can both be true.

index.bs
```diff
@@ -837,11 +837,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 					1. Abort these steps.
 
-				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the sanitized copy of text/plain format.
+				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the content of text/plain format.
```
BoCupp-Microsoft, line 840: Suggest that we change:

>For the MIME types defined in the [=mandatory data types=] list, |data| contains the content of text/plain format.

to 

For the MIME types defined in the [=mandatory data types=] list, |data| MAY be sanitized.

index.bs
```diff
@@ -837,11 +837,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
 
 					Issue: Some OSs contain multiple clipboard (e.g. Linux, "primary", "secondary", "selection"). Define from which of those data is read.
 
-					Issue: Add definition of sanitized copy.
+					Issue: Currently Safari does some sanitization. See https://github.com/w3c/clipboard-apis/issues/150.
```
BoCupp-Microsoft, line 844: Not sure if we need this issue if we have agreement on saying data MAY be sanitized above.

index.bs
```diff
@@ -953,17 +953,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 							1. If |type| is not in the [=mandatory data types=] list, then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm| and abort these steps.
 
-							1. Let |cleanItem| be a sanitized copy of |blob|.
```
BoCupp-Microsoft, line 956: Maybe instead change to be:

1. Let |cleanItem| be an optionally sanitized copy of |blob|

index.bs
```diff
@@ -953,17 +953,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
-
-								Issue: Add definition of sanitized copy.
-
-							1. If unable to create a sanitized copy, then follow the below steps:
```
BoCupp-Microsoft, line 960: Maybe change to be:

1. If sanitization was attempted and was not successfully completed, then follow the below steps:

index.bs
```diff
@@ -1702,19 +1694,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 		1. Else
 
-			1. This is left to the implementation...
-
-				Issue: It's not good to leave things up to the
-				implementation. What should happen here?
+			1. Assign "" to |wellKnownFormat|.
```
BoCupp-Microsoft, line 1697: This seems like it could be a separate PR as it isn't directly related to allowing an optional sanitization of the well-known formats?

index.bs
```diff
@@ -786,15 +786,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 					1. Abort these steps.
 
-				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the sanitized copy of text/html format, but image/png format has unsanitized payload to preserve meta data.
+				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| contains the content of text/html format, but image/png format has unsanitized payload to preserve meta data.
```
BoCupp-Microsoft, line 789: Suggest that we change the sentence starting with:

> For the MIME types defined...

to 

For the MIME types defined in the [=mandatory data types=] list, |data| MAY be sanitized

index.bs
```diff
@@ -786,15 +786,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
 
 					Issue: Some OSs contain multiple clipboard (e.g. Linux, "primary", "secondary", "selection"). Define from which of those data is read.
 
+					Issue: Currently Safari does some sanitization for [=mandatory data types=]. See https://github.com/w3c/clipboard-apis/issues/150.
```
BoCupp-Microsoft, line 793: Same as I mentioned below... not sure if we need this called out if we are using language that makes it clear sanitization is optional.

BoCupp-Microsoft:
> Yes and @whsieh and I argued for allowing sanitization at least for built-in types too. Those can both be true.

@annevk Yes we agree that sanitization can be allowed for built-in types.  I thought that we were in agreement also that it didn't need to be required.  Maybe the language I suggested in my review of this PR can help resolve the issue?

annevk:
@BoCupp-Microsoft that's correct. Sanitization would be optional and the same for legacy/new APIs (ideally through them using a common algorithm).

index.bs
snianu: Removed.

index.bs
snianu: Done.

index.bs
snianu: Removed.

index.bs
snianu: Done.

index.bs
snianu: Done.

index.bs
snianu: Done.

index.bs
```diff
@@ -867,12 +867,16 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 										1. Resolve |p| with |v|.
 
+										1. Return |p|.
```
BoCupp-Microsoft, line 870: Let's just make one change at a time focused on describing optional sanitization.  Could you remove these new returns and other editorial changes?

BoCupp-Microsoft (approved):

Issue #167, opened 2022-02-02
Clipboard Pickling: should the unsanitized option support a way to ask for all representations to be unsanitized?


BoCupp-Microsoft:
Mostly authors are only interested in reading representations from the clipboard that they know how to ingest into their app.  For building a clipboard examiner app [like this](http://whsieh.github.io/examples/datatransfer), however, you don't know what types you need to request until after you read.

Since the [current proposal](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md) requires the author to explicitly list every custom mime-type before they call read, it seems like a the clipboard examiner app won't be able to show  everything that could possibly be requested from the clipboard with the current proposal.

There was an [alternative considered](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#alternative-considered-unsanitizedtrue) for `unsanitized:true`  in the explainer that said it would cause all representations to be read unsanitized.  That could solve the problem, but it would prevent the author from simultaneously requesting a sanitized copy of the well-known representations if they that's what they wanted.

Another option may be to move the unsanitized option to the `ClipboardItem.getType` call instead of the `Clipboard.read`.

@snianu

snianu:
@mbrodesser @annevk

annevk:
I see where the `ClipboardItem.prototype.types` question in the other issue comes from now, I think. It would make sense to me if that lists all the types (lexicographically sorted) that the `ClipboardItem` contains. Otherwise you don't know what you can `getType()`.

snianu:
@annevk Agreed that `ClipboardItem.prototype.types` should list all the types(both custom & well known formas), but not sure about sorting lexicographically as there are platform specific requirements where the most desired format should be listed at the top. In this case, it would be custom formats at the top, well-known formats come after.

annevk:
If platforms have a deterministic sorting order for built-in formats that works too, but presumably we need to say something for custom formats?

PR #175, opened 2022-05-05
Add Web Custom format

### LINKED ISSUE #165: Arbitrary clipboard types
As the current async clipboard API does not have a definition, extensions to it are somewhat hard to define, but here is my basic idea for arbitrary clipboard types that I think works for all parties.

The `ClipboardItem` constructor and the `ClipboardItem` `getType()` method both take a string that represents a type. As this type is a MIME type it's formatted as "`type/subtype`". The only types supported there are builtin types, such as "`text/plain`".

To extend this to allow arbitrary types, without impacting the processing of builtin types, which can be special depending on the user agent and its requirements, this proposal suggests introducing a namespace for arbitrary types. Arbitrary types will always have their data stored and read as-is, without any manipulation.

Arbitrary types are denoted by the prefix "`web `" ("`web`" followed by U+0020 SPACE) to distinguish them from builtin types. They are stored in an implementation-defined manner on the underlying OS clipboard and must not interfere with builtin types.

Expressed in JavaScript, this is how would you read an arbitrary type:

```js
const items = await navigator.clipboard.read();
const arbitraryBlob = await items[0].getType(`web ${arbitraryType}`);
```

And this is how would write one:

```js
const items = [new ClipboardItem({ `web {$arbitraryType}`: arbitraryBlob })];
navigator.clipboard.write(items);
```

@whsieh @snianu @BoCupp-Microsoft @travisleithead @johanneswilm

snianu:
Closes #165 

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [X] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/pull/175.html" title="Last updated on Sep 18, 2023, 6:34 PM UTC (c6394d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/175/4db562e...c6394d3.html" title="Last updated on Sep 18, 2023, 6:34 PM UTC (c6394d3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -165,15 +165,11 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 	The platform provides a <dfn>system clipboard</dfn>.
 
-	The [=system clipboard=] has a list of clipboard items that are collectively
-	called the <dfn>system clipboard data</dfn>.
+	The [=system clipboard=] has a list of [=system clipboard item=]s that are collectively	called the <dfn>system clipboard data</dfn>.
 
-	* For the [[#clipboard-event-api]], the [=system clipboard data=] is
-		exposed as a {{DataTransfer}} object that mirrors the contents of the clipboard.
+	Each <dfn>system clipboard item</dfn> has a list of [=system representations=].
 
-	* For the [[#async-clipboard-api]], the [=system clipboard data=] is
-		exposed as a [=list of representations=] that mirrors the contents of
-		the clipboard.
+	Each <dfn>system representation</dfn> has a <dfn for="system representation">name</dfn>, which is a string, and <dfn for="system representation">data</dfn>, which is a sequence of bytes.
 
 <h2 id="clipboard-events-and-interfaces">Clipboard Events</h2>
 
@@ -634,7 +630,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			<dd>Returns true if <var>type</var> is in [=mandatory data types=] or [=optional data types=], else, returns false.</dd>
 		</dl>
 
-		A <dfn>clipboard item</dfn> is conceptually data that the user has expressed a desire to make shareable by invoking a "cut" or "copy" command. A [=/clipboard item=] serves two purposes. First, it allows a website to read data copied by a user to the system clipboard. Second, it allows a website to write data to the system clipboard.
+		A <dfn>clipboard item</dfn> is conceptually data that the user has expressed a desire to make shareable by invoking a "cut" or "copy" command. A [=/clipboard item=] serves two purposes. First, it allows a website to read data copied by a user to the [=system clipboard=]. Second, it allows a website to write data to the [=system clipboard=].
 
 		<p class=note>
 			For example, if a user copies a range of cells from a spreadsheet of a native application, it will result in one [=/clipboard item=]. If a user copies a set of files from their desktop, that list of files will be represented by multiple [=/clipboard item=]s.
@@ -642,7 +638,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 		Some platforms may support having more than one [=/clipboard item=] at a time on the [=clipboard=], while other platforms replace the previous [=/clipboard item=] with the new one.
 
-		A [=/clipboard item=] has a <dfn>list of representations</dfn>, each <dfn>representation</dfn> with an associated <dfn for=representation>mime type</dfn> (a [=/MIME type=]) and <dfn for=representation>data</dfn> (a {{ClipboardItemData}}).
+		A [=/clipboard item=] has a <dfn>list of representations</dfn>, each <dfn>representation</dfn> with an associated <dfn for=representation>mime type</dfn> (a [=/MIME type=]), an <dfn for="representation">is custom</dfn> flag, initially |false|, that indicates if this [=representation=] should be treated as a [=web custom format=] (as opposed to a well-known format of the [=system clipboard=]), and <dfn for=representation>data</dfn> (a {{ClipboardItemData}}).
 
 		<p class=note>
 			In the example where the user copies a range of cells from a spreadsheet, it may be represented as an image (image/png), an HTML table (text/html), or plain text (text/plain).
@@ -688,17 +684,25 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			1. For each (|key|, |value|) in |items|:
 
-				1. Let |webCustomFormatKey| be an empty string.
+				1. Let |representation| be a new [=representation=].
 
-				1. If |key| has "web "("web" followed by U+0020 SPACE) prefix, then remove "web " prefix and set the remaining string to |webCustomFormatKey|.
+				1. Let |isCustom| be |false|.
+				
+				1. If |key| has "web " ("web" followed by U+0020 SPACE) prefix, then 
+				
+					1. Remove "web " prefix and set the remaining string to |key|.
 
-				1. If |webCustomFormatKey| is not empty, then let |mimeType| be the result of [=parsing a MIME type=] given |webCustomFormatKey|, else let |mimeType| be the result of [=parsing a MIME type=] given |key|.
+					1. Set |isCustom| |true|
+					
+				1. Set |representation|'s [=representation/is custom=] flag to |isCustom|.
+
+				1. Let |mimeType| be the result of [=parsing a MIME type=] given |key|.
 
 				1. If |mimeType| is failure, then throw a {{TypeError}}.
 
-				1. If [=this=]'s [=ClipboardItem/clipboard item=]'s [=list of representations=] [=list/contains=] a [=representation=] whose [=representation/MIME type=] is |mimeType|, then throw a {{TypeError}}. 
+				1. If [=this=]'s [=ClipboardItem/clipboard item=]'s [=list of representations=] [=list/contains=] a [=representation=] whose [=representation/MIME type=] is |mimeType| and whose [representation/is custom] is |isCustom|, then throw a {{TypeError}}.
 
-				1. Let |representation| be a new [=representation=].
+				<p class=note>The step above prevents collision between mime-types well-known by the user agent and those that are intended by the author to be treated as a custom type.  For example, it would be possible for the author's list of items to contain representations for "text/html" as well as "web text/html".</p>
 
 				1. Set |representation|'s [=representation/MIME type=] to |mimeType|.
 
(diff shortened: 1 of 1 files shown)


snianu:
@annevk @mbrodesser This PR has the web prefix changes. PTAL. Thanks!

annevk:
Upon doing code inspection in Gecko we found that the legacy clipboard API might have some support for custom types as well. @EdgarChen is investigating this. We don't really want to end up with a situation where we have two incompatible custom types APIs.

It might be that other browsers have something similar as there were at least some hints of cross-browser collaboration?

snianu:
@annevk Chromium and Safari have browser specific custom type that is undocumented and doesn't interop with other apps. The format parsing rules are convoluted since it uses browser specific pickling types for serialization/deserialization. This proposal aims to standardize the web custom formats and the format structure so its well documented and is interop across apps on a platform. This doesn't affect the existing pickling types.

In Firefox, I see that it allows writing custom types in setData, but I don't see any restriction on the number of custom types that can be registered. This is a major security issue on Windows and Linux because of the restriction on the number of clipboard formats that can be registered with the OS. This proposal addresses this issue via a fixed set of web custom formats and a map that maps the mime types to native web custom formats.
image

index.bs
```diff
@@ -786,7 +786,7 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 					1. Abort these steps.
 
-				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| MAY be sanitized, but image/png format has unsanitized payload to preserve meta data.
+				1. Let |data| be a copy of the [=system clipboard data=] represented as [=clipboard items=]. For the MIME types defined in the [=mandatory data types=] list, |data| MAY be sanitized, but image/png format has unsanitized payload to preserve meta data. For the [=/MIME type=]s that have a "web "("web" followed by U+0020 SPACE) prefix, [=read unsanitized format=] from the [=system clipboard=] and append the content of the format to |data|.
```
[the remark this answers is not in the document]
snianu: Added these additional steps.

index.bs
```diff
@@ -1638,83 +1646,161 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
-				1. Return |wellKnownFormat|.
+		1. Return |wellKnownFormat|.
+
+	</div><!-- algorithm -->
+
+	<div class="algorithm" data-algorithm="read-unsanitized-format">
+	<h3 id="to-read-unsanitized-format"><dfn>read unsanitized format</dfn></h3>
+
+		: Input
+		:: |mimeType|, a string
```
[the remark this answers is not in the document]
snianu: To be a bit more resilient to the parsing mime type errors, I added steps to ignore the mime types that returned failure during parsing.

index.bs
```diff
@@ -1638,83 +1646,161 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
+
+		1. Let |webCustomFormatMap| be the [=os specific custom map name=].
+
+		1. Read |webCustomFormatMap| from the [=system clipboard=].
+
+		1. Let |webCustomFormatMapString| be the JSON string deserialized from |webCustomFormatMap|.
+
+			Note: Need a JSON reader to deserialize the content from the |webCustomFormatMap|.
+
+		1. Let |webCustomFormat| be the value corresponding to the |mimeType| key in |webCustomFormatMapString|.
```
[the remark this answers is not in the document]
snianu: I'm looping through the key value pair now, so removed this extra step as it's redundant.

index.bs
```diff
@@ -1638,83 +1646,161 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
+
+				Note: Need a JSON writer to serialize the content into the |webCustomFormatMap|.
+
+			1. Insert the |item| into the [=system clipboard=] using |webCustomFormatIdx| as the format.
+
+			1. Increment |idx|.
+
+			1. If |idx| is greater than 100, then break from this loop.
+
+		1. Insert the |webCustomFormatMap| into the [=system clipboard=].
```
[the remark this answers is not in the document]
snianu: Done.

index.bs
```diff
@@ -1638,83 +1646,161 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
+
+		1. Let |webCustomFormatMap| be the [=os specific custom map name=].
+
+		1. For each |item| in |items|:
+
+			1. Let |webCustomFormat| be the [=os specific custom name=].
+
+			1. Let |webCustomFormatIdx| be the result of appending |idx| to |webCustomFormat|.
+
+			1. Insert |mimeType| as key and |webCustomFormatIdx| as value into the |webCustomFormatMap|.
```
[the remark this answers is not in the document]
snianu: Done.

EdgarChen:
> This proposal aims to standardize the web custom formats and the format structure so its well documented and is interop across apps on a platform. This doesn't affect the existing pickling types.

Do we care about the interop across legacy clipboard API and async clipboard API for custom type? E.g. write custom type via legacy clipboard API and read via async clipboard API, or vice versa.

snianu:
> Do we care about the interop across legacy clipboard API and async clipboard API for custom type?

The async clipboard API model is fairly different, so I'm not sure if we can interop with legacy types. The web custom types are stored inside a JSON map structure, with fixed number of native web clipboard formats, and a well-defined format structure that helps the native apps parse the content of the format. At least in Chromium, the parsing of the custom types requires browser specific pickle types to serialize and deserialize the content, so it's not possible to add support for this format to the new web custom format.

snianu:
@EdgarChen While it's true that you could write custom types with "web " prefix using DataTransfer APIs, those custom types do not follow the same model - there isn't a restriction on the number of custom format types that can be written by the web authors, no web custom JSON map for mapping the web custom types to a fixed number of native web clipboard formats etc.
We don't want to make any changes in the legacy DataTransfer APIs as it has high regression risks and the model is fairly different there(at least in Chromium). The native apps on Windows do not know how to parse Chromium specific custom types, so it's not possible for them to exchange formats between web and native apps.
Chrome doesn't expose the custom mime types exactly as is - it exposes it via a Chromium specific custom format that uses browser specific parsing types, and it is not standardized, so the native apps don't understand how to parse this structure.

index.bs
```diff
@@ -786,26 +805,60 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
...
+
+						1. Append |representation| to |item|'s [=list of representations=].
+					
+					1. If |item|'s [=list of representations=] size is greater than 0, append |item| to |items|.
+					
+				If |items| has a size > 0, then:
+				
+					1. Let |firstItem| be |items|[0]
+					
+					1. Run the [=read web custom format=] algorithm given |firstItem|.
```
EdgarChen, line 840: Per https://w3c.github.io/clipboard-apis/#dom-clipboarditem-types, `type` getter of ClipboardItem returns `types array`, but it is not clear how does `types array` be constructed. It would be nice that we could have some steps for that.

Not sure if this makes sense: maybe we could construct `types array` from ` list of representations`, but which means we would need to read custom types first as [we would like the custom types listed at the top](https://github.com/w3c/clipboard-apis/issues/167#issuecomment-1031748387).

snianu: I agree. We should definitely add this to the spec.

polx:
Hello @snianu ,

I've tried to review the current state after I have seen your presentation at the TPAC.

Here are my rather nitty gritty comments:

- sec 7.2 (ClipboardItem Interface): should `is custom` really be written in two words and not in one? The link in the version I could build did not work but that might be a local issue.
- the spreadsheet example close to that is nice, could it not also include a custom type? (e.g. containing  text/csv alternative?)
- A web custom format has is custom set to true. => A representation in a web custom format has is custom set to true ?
- 7.3.3: write algorithm: f type is not in the mandatory data types list and has "web ":   did you not mean “or has...” ?
- write blobs and option to the clipboard:
   - step 1: why is the variable called webCustomFormat ? It should be formats or something universal, right?
   - substeps under 2.2: "If formatString is empty then follow the below steps:": the steps here does not define a (non-empty) formatString; a value is however needed in order for the recipients to read (suggestion: “set formatString to  a randomly chosen format name”)
  - step 2.3: Let payload be the result of UTF-8 decoding item’s underlying byte sequence. : this must be a misunderstanding as it would only apply to text. Maybe the whole section was about text? Then it was not readable
- the steps of “write web custom formats” sound quite solid to me but why is it independent from the previous?

Tell me if I should make my comments more precise (e.g. with line numbers).

I am quite excited by the implementation found in Chromium (and [blogged about it](https://mathmlmuses.netlify.app/trymathml-copy-customtypes/)). I suspect that we are getting close to acceptance, at least provided that Mozilla will consider implementing as it seems it said at the start.

thanks and sorry to be slow to have reviewed.

Paul

snianu:
Thanks @polx for the review and apologies for the delay!
> A web custom format has is custom set to true. => A representation in a web custom format has is custom set to true ?

Note that in the spec it's referred to as `[=representation/isCustom=]`

> step 1: why is the variable called webCustomFormat ? It should be formats or something universal, right?

This is referring to the custom format with a `web ` prefix, so it can be anything. In the below step, we strip out the `web ` prefix:
`If |webCustomFormatString| has "web "("web" followed by U+0020 SPACE) prefix, then remove the "web " prefix and store the remaining string in |webCustomFormat|'s {{Blob/type}}'s [=MIME type/essence=], else abort all steps.`

> step 2.3: Let payload be the result of UTF-8 decoding item’s underlying byte sequence. : this must be a misunderstanding as it would only apply to text. Maybe the whole section was about text? Then it was not readable

This is by-design that we write raw bytes sequence to the clipboard for the custom formats so the reader has to know how to parse its content (which is also by-design).

> the steps of “write web custom formats” sound quite solid to me but why is it independent from the previous?

This is because we need to perform an additional step to produce a native clipbaord format and insert the key-value pair into the `webCustomFormatMap`. This step is not performed for mandatory data types.

snianu:
@sanketj could you please take a look at this PR so we can merge it into the clipboard repo? Thanks!

garykac (approved):

snianu:
@garykac Thank you so much!

index.bs
```diff
@@ -688,13 +685,25 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			1. For each (|key|, |value|) in |items|:
 
+				1. Let |representation| be a new [=representation=].
+
+				1. Let |isCustom| be |false|.
+				
+				1. If |key| has "web " ("web" followed by U+0020 SPACE) prefix, then
```
evanstade, line 692: Is there a "starts with" definition for strings we can reference here?

mkruisselbrink: https://infra.spec.whatwg.org/#string-starts-with

snianu: Will make this change, thanks!

snianu: https://github.com/w3c/clipboard-apis/pull/195

PR #192, opened 2023-09-14
Add supports API for async clipboard.

### LINKED ISSUE #170: Feature detection for supported clipboard formats
Migrated and adapted from https://github.com/w3c/clipboard-apis/issues/165#issuecomment-1041668421:

I suggest we add some sort of feature-detection for custom and classic formats. This is actual nightmare-ish application code of mine that only makes sure copying works on three browsers (Firefox, Chrome, Safari): https://github.com/tomayac/SVGcode/blob/b29b27dadf2cf3c8aefb77e2c3db9feb0872d91d/src/js/clipboard.js#L64-L136

My suggestion would be to add a static function on `ClipboardItem`

```js
ClipboardItem.supports('image/webp');
// `false`
ClipboardItem.supports('image/png');
// `true`
ClipboardItem.supports('web');
// `true`
```

Not sure how to express custom types here, as a ground for discussion I just went with `"web"`. What do you think?

snianu:
Closes #170 

 * [X] Modified Web platform tests (link to pull request)

Implementation commitment:

 * [X] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [X] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -580,6 +580,8 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 				readonly attribute FrozenArray&lt;DOMString> types;
 
 				Promise&lt;Blob&gt; getType(DOMString type);
+
+				static boolean supports(DOMString type);
 			};
 
 			enum PresentationStyle { "unspecified", "inline", "attachment" };
@@ -608,6 +610,9 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			<dt><code><var>clipboardItem</var>.<var>types</var></code></dt>
 			<dd>Returns the list of [=representation/mime type=]s contained in the [=/clipboard item=] object.</dd>
+
+			<dt><code>{{ClipboardItem}}.supports(<var>type</var>)</code></dt>
+			<dd>Returns true if <var>type</var> is in [=mandatory data types=], else, returns false.</dd>
 		</dl>
 
 		A <dfn>clipboard item</dfn> is conceptually data that the user has expressed a desire to make shareable by invoking a "cut" or "copy" command. A [=/clipboard item=] serves two purposes. First, it allows a website to read data copied by a user to the system clipboard. Second, it allows a website to write data to the system clipboard.
@@ -741,6 +746,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 			1. [=Reject=] |p| with {{"NotFoundError"}} {{DOMException}} in |realm|.
 
 			1. Return |p|.
+
+		<h4 method for=ClipboardItem lt=supports(type)>supports(type)</h4>
+
+			This method must run the below steps:
+
+			1. If |type| is in [=mandatory data types=], then return true.
+
+			1. If |type| is not in [=mandatory data types=], then return false.
+
 
 	<h3 id="clipboard-interface">Clipboard Interface</h3>
 


evanstade:
@garykac could you ptal?

index.bs
```diff
@@ -742,6 +747,15 @@ urlPrefix: https://w3c.github.io/FileAPI/#dfn-; type: dfn;
 
 			1. Return |p|.
 
+		<h4 method for=ClipboardItem lt=supports(type)>supports(type)</h4>
+
+			This method must run the below steps:
+
+			1. If |type| is in [=mandatory data types=], then return true.
+
+			1. If |type| is not in [=mandatory data types=], then return false.
```
saschanaz, line 756: Maybe this can be just "If not"?

BTW, does this mean the API supports only "mandatory data types" and not anything else?

sanketj (commented):
In addition to the mandatory data types, should this API also return true for correctly-prefixed custom formats?

snianu:
@sanketj Yes, I'm planning to add that in PR https://github.com/w3c/clipboard-apis/pull/175

index.bs
snianu: That is a good point. We [removed the types](https://github.com/w3c/clipboard-apis/pull/155) that may be supported by a user agent, but I think it would be good to have that list as we're adding this new API to check supported formats. I'll add it in a separate section. Thanks for reviewing this PR!

sanketj (approved):

snianu:
@evanstade @garykac Note that this feature (including the API shape) was[ approved by EditingWG](https://github.com/w3c/clipboard-apis/issues/170#issuecomment-1099368334) already.

evanstade:
I think the exact wording and behavior here still warrant discussion even though the EditingWG approved the general idea of it (?) 1.5 years ago.

Specifically, which of the following should return true if web custom formats are supported, and why?

```
ClipboardItem.supports('web');  // 1
ClipboardItem.supports('web ');  // 2
ClipboardItem.supports('web image/jpeg');  // 3
ClipboardItem.supports('web text/plain');  // 4
ClipboardItem.supports('web txt/plain');  // 5
ClipboardItem.supports('web application/foobar');  // 6
```
1,2. These are not valid types per se, just a part of a type. But if a site just wants to know if web custom formats are supported, then one of these is probably sufficient. That is, the browser only needs to return true for the exact string `web` or `web `.

4. Is there a point in supporting this? Is it equivalent to `text/plain` alone, or not?
5. This is not a real MIME type. Should it be "supported"?
6. This is a real MIME type.

snianu:
@evanstade Per [the web custom format proposal](https://github.com/w3c/clipboard-apis/issues/165#issue-1117218251), web custom formats are arbitrary types that are denoted by the prefix "web " ("web" followed by U+0020 SPACE) to distinguish them from built-in types, so we only need to check for the web prefix. The custom MIME type is specified by the web author at the time of ClipboardItem object creation which is when we validate the `type/subtype` format. See [step 5 in the constructor ](https://w3c.github.io/clipboard-apis/#dom-clipboarditem-clipboarditem)where this validation is done.
So, in your example we will return true for all cases except 1.

evanstade:
In the original request for `supports`, the exact opposite is suggested: https://github.com/w3c/clipboard-apis/issues/165#issuecomment-1041668421

And I think that makes more sense. What is the purpose of any of the text after "web"? It has no bearing on the returned value.

snianu:
I didn't notice that the proposal doesn't have a space after `web` prefix. It's essential as it differentiates between web custom formats and other valid MIME type formats that browsers may not have support for in the async clipboard write method. e.g. `web text/html` is supported but `webtext/html` is not.

evanstade:
I don't think that `webtext/html` is actually a valid MIME type. There are only [8 valid types](https://www.iana.org/assignments/media-types-parameters/media-types-parameters.xhtml) (that is, the part before the slash).

Either way, the proposal is for an exact string match on `web`.

snianu:
I was referring to [this ](https://mimesniff.spec.whatwg.org/#mime-type)definition of MIME type which is what we use in the clipboard spec. `webtext/html` passes[ the parsing check of the MIME types](https://mimesniff.spec.whatwg.org/#parse-a-mime-type), so from a clipboard perspective, it's a valid MIME type. It's just that we don't have support for it in the clipboard.

IMO removing the space from the `web ` prefix would create confusion among web authors, and since it's part of how we [parse the web custom format](https://w3c.github.io/clipboard-apis/#dom-clipboarditem-clipboarditem) during the construction of the ClipboardItem, our preference is to match with the `web ` prefix ("web" followed by U+0020 SPACE).

Anyways, I don't have a strong opinion on this so I'm adding Agenda+ label to discuss with the EditingWG members.

snianu:
Note that currently web authors already use `startsWith` and `web ` to check if there is any web custom format after read operation: https://developer.chrome.com/blog/web-custom-formats-for-the-async-clipboard-api/#reading-web-custom-formats-from-the-clipboard.
So, the current proposal of checking for `web ` aligns with this approach as well.

evanstade:
If you are given a string and you have to figure out whether it represents a web custom type, then sure, use `startsWith` and `web `. But what we are discussing is: what is the best argument to represent the concept of "web custom formats in general"? We're talking about a string only because the argument is also used to reference specific built-in types, for which a string is the most appropriate type.

I'd be interested in hearing opinions from folks with long-term spec-writing experience such as @inexorabletash

inexorabletash:
I don't have strong feelings here, but my intuition for how developers would try to use supports() for custom types in practice is either:

* `ClipboardItem.supports('web valid/type');` // where we'd return true for 'web' + space prefix and valid mime type suffix suffix
* `ClipboardItem.supports('web');` // true

The former aligns with the concept of supports() being a way of testing what the ClipboardItem constructor would do without wasting time, but accepting 'web' on its own seems like a nice to have. 

'web ' (web + space, with no suffix) on its own isn't accepted by https://w3c.github.io/clipboard-apis/#dom-clipboarditem-clipboarditem so far as I can tell (maybe I'm wrong?), and it's not intuitive to me that a developer would try it.

But maybe I'm missing the point of the discussion. What are the alternatives we're considering?

snianu:
> ClipboardItem.supports('web valid/type'); // where we'd return true for 'web' + space prefix and valid mime type suffix suffix

I like this proposal as it also matches with how we parse the web custom format during the `ClipboardItem` [construction](https://w3c.github.io/clipboard-apis/#dom-clipboarditem-clipboarditem). This also eliminates confusion between `web text/html` and `webtext/html`.
Tagging few other people who were involved in the initial discussion of this proposal for feedback. @annevk @whsieh @EdgarChen @tomayac

tomayac:
I'm happy with https://github.com/w3c/clipboard-apis/pull/192#issuecomment-1753661815, it extends https://github.com/w3c/clipboard-apis/issues/170#issue-1140281174 in a very logical way.

evanstade:
> * `ClipboardItem.supports('web valid/type');` // where we'd return true for 'web' + space prefix and valid mime type suffix suffix

I suppose it is the case that the part after `'web '` has to be a valid MIME type. I missed that detail because it's not mentioned under [optional data types](https://w3c.github.io/clipboard-apis/#optional-data-types-x). Perhaps that section should be updated to indicate that the prefix alone is not sufficient, and there are requirements for the rest of the string/type?

> But maybe I'm missing the point of the discussion. What are the alternatives we're considering?

Which of [these](https://github.com/w3c/clipboard-apis/pull/192#issuecomment-1726703751) should return true?

The reason I questioned the utility of `supports('web image/foo')` is because, as far as I understood, everything after the `'web'` prefix didn't matter. If the API accepts full custom types like `web text/richtext`, that suggests a user agent might return a different value for one full custom type vs another, but if the implementation is "just check the first 4 characters" then that's a false impression. Given that the suffix does at least have to be a valid MIME type (`'web foo/bar'`) should return false, it does matter a little bit so it does make more sense to me that the API can take full custom types.

snianu:
The PR for the resolution is https://github.com/w3c/clipboard-apis/pull/195

PR #197, opened 2023-10-30
Add `unsanitized` option to async clipboard API.

snianu:
Closes #????

 * [X] Modified Web platform tests (link to pull request) https://github.com/web-platform-tests/wpt/pull/35528

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1268679)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/pull/197.html" title="Last updated on Nov 18, 2023, 1:29 AM UTC (0312e3e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/197/ad65071...0312e3e.html" title="Last updated on Nov 18, 2023, 1:29 AM UTC (0312e3e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -544,17 +544,9 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
 
-		The implementation MAY recognize the native OS clipboard format description
-		for the following data types, to be able to populate the
-		{{ClipboardItem}} with the correct [=/MIME type=], and
-		set the correct data format on the OS clipboard in response to copy and cut
-		events.
+			These data types MUST NOT be sanitized by UAs:
 
-			These data types must not be sanitized and should be exposed by UAs
-			if a corresponding native type exists on the clipboard:
-
-			* text/html
-
+			* image/png (which should remain unsanitized to preserve meta data)
 
 <h2 id="async-clipboard-api">Asynchronous Clipboard API</h2>
 
@@ -816,7 +808,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
-		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are in the [=unsanitized data types=].
+		<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST not be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
+
+			* text/html
+
+		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are [=optional unsanitized data types=].
 
 		The <dfn>clipboard task source</dfn> is triggered in response to reading or writing of [=system clipboard data=].
 
@@ -833,7 +829,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 				1. For each |format| in |formats|["{{ClipboardUnsanitizedFormats/unsanitized}}"]:
 
-					1. If |format| is not in [=unsanitized data types=], then [=reject=] |p| with |format| {{"NotAllowedError"}} {{DOMException}} in |realm|.
+					1. If |format| is not in [=optional unsanitized data types=], then [=reject=] |p| with |format| {{"NotAllowedError"}} {{DOMException}} in |realm|.
 
 			1. Run the following steps [=in parallel=]:
 
@@ -877,7 +873,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 						
 						1. The user agent, MUST NOT sanitize |representation|'s [=representation/data=], if it satisfies the below conditions:
 						
-							1. |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is "image/png", which should remain unsanitized to preserve meta data.
+							1. |representation|'s [=representation/MIME type=] is in [=unsanitized data types=] list.
 							
 							1. |isUnsanitized| is true.
 
--- a/index.bs
+++ b/index.bs
@@ -546,7 +546,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 			These data types MUST NOT be sanitized by UAs:
 
-			* image/png (which should remain unsanitized to preserve meta data)
+			* image/png
+
+			<p class=note>
+				The image/png MIME type is not sanitized to preserve meta data.
+			</p>
 
 <h2 id="async-clipboard-api">Asynchronous Clipboard API</h2>
 
@@ -808,11 +812,10 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
-		<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST not be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
+		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are [=optional unsanitized data types=].
+		<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST NOT be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
 
 			* text/html
-
-		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are [=optional unsanitized data types=].
 
 		The <dfn>clipboard task source</dfn> is triggered in response to reading or writing of [=system clipboard data=].
 
--- a/index.bs
+++ b/index.bs
@@ -542,6 +542,19 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 			* Custom format [=string/starts with=] `"web "`("web" followed by U+0020 SPACE) prefix
 				and suffix (after stripping out `"web "`) passes the [=parsing a MIME type=] check.
 
+	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
+
+		The implementation MAY recognize the native OS clipboard format description
+		for the following data types, to be able to populate the
+		{{ClipboardItem}} with the correct description for paste events, and
+		set the correct data format on the OS clipboard in response to copy and cut
+		events.
+
+			These data types must not be sanitized and should be exposed by UAs
+			if a corresponding native type exists on the clipboard:
+
+			* text/html
+
 
 <h2 id="async-clipboard-api">Asynchronous Clipboard API</h2>
 
@@ -620,6 +633,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		Some platforms may support having more than one [=/clipboard item=] at a time on the [=clipboard=], while other platforms replace the previous [=/clipboard item=] with the new one.
 
 		A [=/clipboard item=] has a <dfn>list of representations</dfn>, each <dfn>representation</dfn> with an associated <dfn for=representation>mime type</dfn> (a [=/MIME type=]), an <dfn for="representation">isCustom</dfn> flag, initially |false|, that indicates if this [=representation=] should be treated as a [=web custom format=] (as opposed to a well-known format of the [=system clipboard=]), and <dfn for=representation>data</dfn> (a {{ClipboardItemData}}).
+
 		A <dfn>web custom format</dfn> has [=representation/isCustom=] set to |true|.
 
 		<p class=note>
@@ -802,11 +816,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
-		An <dfn>unsanitized</dfn> object is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=]. It contains <dfn>unsanitized mime types</dfn> that is currently limited to "text/html".
-
-		<p class="note">
-		In the future, other MIME types may be added.
-		</p>
+		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are in the [=unsanitized data types=].
 
 		The <dfn>clipboard task source</dfn> is triggered in response to reading or writing of [=system clipboard data=].
 
@@ -819,15 +829,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 			1. Let |p| be [=a new promise=] in |realm|.
 
-			1. Let |format| be a {{DOMString}}.
-
 			1. If |formats| is not empty, then:
 
-					1. If |formats|'s size is greater than 1, then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
+				1. For each |format| in |formats|["{{ClipboardUnsanitizedFormats/unsanitized}}"]:
 
-					1. Set |format| to |formats|[0].
-
-					1. If |format| is not in [=unsanitized mime types=], then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
+					1. If |format| is not in [=unsanitized data types=], then [=reject=] |p| with |format| {{"NotAllowedError"}} {{DOMException}} in |realm|.
 
 			1. Run the following steps [=in parallel=]:
 
@@ -857,15 +863,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 						
 						1. Set |representation|'s [=representation/MIME type=] to |mimeType|.
 
+						1. Let |isUnsanitized| be |false|.
+
+						1. If |formats| is not empty, then:
+
+							1. For each |format| in |formats|["{{ClipboardUnsanitizedFormats/unsanitized}}"]:
+
+								1. If |format| is equal to [=representation/MIME type=], set |isUnsanitized| to true.
+
 						1. Set |representation|'s [=representation/data=] to |systemClipboardRepresentation|'s [=system clipboard representation/data=].
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless it satisfies the below conditions:
+						1. The user agent, MUST NOT sanitize |representation|'s [=representation/data=], if it satisfies the below conditions:
 						
 							1. |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is "image/png", which should remain unsanitized to preserve meta data.
 							
-							1. If |format| is not empty and |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is equal to |format|.
+							1. |isUnsanitized| is true.
 
 						1. Append |representation| to |item|'s [=list of representations=].
 					
--- a/index.bs
+++ b/index.bs
@@ -550,9 +550,9 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 			* [=optional unsanitized data types=]
 
-			<p class=note>
-				The image/png MIME type is not sanitized to preserve meta data.
-			</p>
+			<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST NOT be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
+
+			* text/html
 
 <h2 id="async-clipboard-api">Asynchronous Clipboard API</h2>
 
@@ -815,9 +815,6 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		</p>
 
 		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that the author wants to be treated as [=optional unsanitized data types=].
-		<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST NOT be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
-
-			* text/html
 
 		The <dfn>clipboard task source</dfn> is triggered in response to reading or writing of [=system clipboard data=].
 
--- a/index.bs
+++ b/index.bs
@@ -782,10 +782,14 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 		[SecureContext, Exposed=Window]
 		interface Clipboard : EventTarget {
-			Promise&lt;ClipboardItems&gt; read();
+			Promise&lt;ClipboardItems&gt; read(optional ClipboardUnsanitizedFormats formats = {});
 			Promise&lt;DOMString&gt; readText();
 			Promise&lt;undefined&gt; write(ClipboardItems data);
 			Promise&lt;undefined&gt; writeText(DOMString data);
+		};
+
+		dictionary ClipboardUnsanitizedFormats {
+			sequence&lt;DOMString&gt; unsanitized;
 		};
 		</pre>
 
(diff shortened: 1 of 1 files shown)


snianu:
@evanstade Made spec changes for `unsanitized` option. Please let me know if you have any comments/concerns. Thanks!

yisibl:
I would like the `unsanitized` option to be applied to `image/svg+xml` as well, which would solve the problem of losing `<style>` when copying SVG.  See https://bugs.chromium.org/p/chromium/issues/detail?id=1226150

index.bs
```diff
@@ -798,17 +802,33 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
+		An <dfn>unsanitized</dfn> object is a [=sequence=] of {{DOMString}} corresponding to the [=representation/mime type=].
+
+		<p class="note">
+		Only "text/html" MIME type is supported for [=unsanitized=] object. In the future, other MIME types may be added.
```
sanketj, line 808: Should we create a section for supported MIME types (with only "text/html" in it for now) and reference it from here and the algorithm below? That will avoid replicating the same information in two places.

index.bs
```diff
@@ -841,7 +861,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data.
+						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is "image/png", which should remain unsanitized to preserve meta data, or equal to |format| if |format| is not empty.
```
sanketj, line 864: I don't follow what is meant by ", or equal to |format| if |format| is not empty" and how it relates to the rest of the paragraph. Could you elaborate?

sanketj (changes requested):

index.bs
snianu: It basically means, the user agent may not sanitize the `|representation|'s [=representation/data=]` if `|representation|'s [=representation/MIME type=]'s [=MIME type/essence=]` is equal to `|format|.`

snianu:
Tagging @EdgarChen in this PR as well. Could you please take a look if you are OK with this change? Thanks!

index.bs
snianu: Done.

index.bs
```diff
@@ -798,17 +802,33 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 
+			1. Let |format| be a {{DOMString}}.
+
+			1. If |formats| is not empty, then:
+
+					1. If |formats|'s size is greater than 1, then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
+
+					1. Set |format| to |formats|[0].
+
+					1. If |format| is not [=unsanitized mime types=], then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
```
sanketj, line 830:
```suggestion
					1. If |format| is not in [=unsanitized mime types=], then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
```

index.bs
sanketj: Hmm not sure I follow. I understood that the UA must not sanitize for "image/png". Could you give me an example of what you're going for with the second clause?

sanketj (changes requested):

index.bs
snianu: Done.

snianu:
https://github.com/w3c/clipboard-apis/pull/197#discussion_r1378176413

Clarified in the latest iteration.

index.bs
```diff
@@ -798,17 +802,33 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
-		<h4 method for="Clipboard">read()</h4>
-		The {{Clipboard/read()}} method must run these steps:
+		<h4 method for="Clipboard">read(|formats|)</h4>
+		The {{Clipboard/read(formats)}} method must run these steps:
 
 			1. Let |realm| be [=this=]'s [=relevant realm=].
 
 			1. Let |p| be [=a new promise=] in |realm|.
 
+			1. Let |format| be a {{DOMString}}.
```
sanketj, line 822: Shouldn't |formats| be a dictionary instead of just a DOMString?

index.bs
```diff
@@ -841,7 +861,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data.
+						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless it satisfies the below conditions:
+						
+							1. |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is "image/png", which should remain unsanitized to preserve meta data.
+							
+							1. If |format| is not empty and |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is equal to |format|.
```
sanketj, line 868: Is this trying to say that the representation's MIME type is amongst the formats inside the sequence of `DOMString`s, which is named `unsanitized`? https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/clipboard/clipboard.idl;l=1?q=clipboard.idl&sq=&ss=chromium

sanketj (changes requested):

index.bs
snianu: Yep, `|formats` is a dictionary, this variable just stores the first DOMString from that dictionary in the below step:
`Set |format| to |formats|[0].`

index.bs
snianu: basically yes. We extract the first DOMString in the `If |formats| is not empty, then` steps and then try to match the MIME type with it in this highlighted step. In this case it will be just `text/html` mime type.

index.bs
evanstade: but you would be iterating over `formats.unsanitized`, rather than `formats`, no?

index.bs
```diff
@@ -782,11 +782,15 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 		[SecureContext, Exposed=Window]
 		interface Clipboard : EventTarget {
-			Promise&lt;ClipboardItems&gt; read();
+			Promise&lt;ClipboardItems&gt; read(optional ClipboardUnsanitizedFormats formats = {});
```
evanstade, line 785: Why is this a dictionary with a single member which is an array?

If this is a dictionary simply because it makes it easier to extend with more parameters later, then I'd have expected the dictionary to be called `readOptions` or something, with an array called `unsanitizedFormats`.

index.bs
```diff
@@ -798,17 +802,33 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 
 			1. Let |realm| be [=this=]'s [=relevant realm=].
 
 			1. Let |p| be [=a new promise=] in |realm|.
 
+			1. Let |format| be a {{DOMString}}.
+
+			1. If |formats| is not empty, then:
+
+					1. If |formats|'s size is greater than 1, then [=reject=] |p| with {{"NotAllowedError"}} {{DOMException}} in |realm|.
```
evanstade, line 826: Feels odd to me that the IDL is written to support more than one unsanitized format, but the algorithm wording is only structured to support exactly one sanitized format. Why not iterate over the whole list and reject if any of them are not in the allowed-to-be-sanitized list?

Also if more formats were added in the future, the set of permittable formats could vary by browser. How are sites to know which formats they can ask for? Is throwing an error the best way to deal with them asking for the wrong formats to be unsanitized, or should some format values just be ignored? If throwing an error, how does the site figure out which format(s) were in error?

I don't have a ton of experience with spec language/algos. Can you find a spec mentor to help with these questions?

index.bs
```diff
@@ -841,7 +861,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data.
+						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless it satisfies the below conditions:
```
evanstade, line 864: It seems like "sanitize" should be defined.

evanstade: We don't have to necessarily define the exact steps that constitute hyopthetical sanitization, but it does seem that it would be useful to describe what it's not, because another way to word this text is: "The user agent MUST NOT sanitize the data if it satisfies the below conditions". So what exactly is the user agent not allowed to do, or must it do, if it's providing "unsanitized" data?

In particular, this came up in [code review](https://chromium-review.googlesource.com/c/chromium/src/+/4779949) where, after some discussion, it was decided that the CFHTML fragment tags should be left in place. The spec should be clear about this so other vendors don't come to a different conclusion at the implementation step.

index.bs
snianu: If we just have `sequence` with `DOMString`, then we cannot make it optional (at least the bindings generator fails in Chromium). If we make it non-optional, then it would break sites that do not want to read unsanitized formats. Dictionary helps us to make this arg optional and `unsanitized` makes it explicit that authors are opting into reading unsanitized formats i.e. when sites call `navigator.clipboard.read({unsanitized: ['text/html']})`, it is clear to the reader that the site is trying to read unsanitized `text/html` format. We got [this feedback ](https://github.com/w3c/editing/issues/315#issue-942457383)from Chrome's security team.

index.bs
snianu: Ah you're right. Missed that part.

index.bs
snianu: Addressed these concerns in the latest iteration. Please let me know if I missed anything. Thanks!

index.bs
snianu: Done.

index.bs
```diff
@@ -542,6 +542,19 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 			* Custom format [=string/starts with=] `"web "`("web" followed by U+0020 SPACE) prefix
 				and suffix (after stripping out `"web "`) passes the [=parsing a MIME type=] check.
 
+	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
+
+		The implementation MAY recognize the native OS clipboard format description
+		for the following data types, to be able to populate the
```
sanketj, line 548: What do you mean by "correct description for paste events"?

index.bs
```diff
@@ -542,6 +542,19 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+		{{ClipboardItem}} with the correct description for paste events, and
+		set the correct data format on the OS clipboard in response to copy and cut
+		events.
+
+			These data types must not be sanitized and should be exposed by UAs
```
sanketj, line 553: Is it right to say "must not be sanitized"? We should only not sanitize if these types are passed as part of the unsanitized option.

index.bs
```diff
@@ -798,17 +816,25 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
+		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are in the [=unsanitized data types=].
```
sanketj, line 819:
```suggestion
		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that should be handled as [=unsanitized data types=].
```

index.bs
```diff
@@ -837,11 +863,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 						
 						1. Set |representation|'s [=representation/MIME type=] to |mimeType|.
 
+						1. Let |isUnsanitized| be |false|.
```
sanketj, line 866: Does this step need to be done for each format (ie. inside the for each loop below)?

index.bs
```diff
@@ -837,11 +863,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+								1. If |format| is equal to [=representation/MIME type=], set |isUnsanitized| to true.
+
 						1. Set |representation|'s [=representation/data=] to |systemClipboardRepresentation|'s [=system clipboard representation/data=].
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data.
+						1. The user agent, MUST NOT sanitize |representation|'s [=representation/data=], if it satisfies the below conditions:
+						
+							1. |representation|'s [=representation/MIME type=]'s [=MIME type/essence=] is "image/png", which should remain unsanitized to preserve meta data.
```
sanketj, line 880: Should we include this condition in the definition of "unsanitized data types" above? Something like: Unsanitized data types are "image/png" and those that are passed by the user as part of the unsanitized option.

sanketj (changes requested):

index.bs
snianu: The intention is to not sanitize any formats that are in the `Unsanitized data types` list. If the web authors specify any other MIME types, then we will reject that MIME type during `read`.

index.bs
snianu: `Unsanitized data types` is a list of MIME types that must not be sanitized. It doesn't describe any processing steps. If the web authors specify these MIME types in the `{{ClipboardUnsanitizedFormats/unsanitized}}`, then we will handle that according to the steps in `read` method.

index.bs
snianu: Ah nice catch. Need to reset it at the end of the loop.

index.bs
snianu: Yeah, this is something that I wasn't sure about. I don't know the spec language for it, but I wanted to write something that allows MIME types in `unsanitized data types` list to be part of `{{ClipboardUnsanitizedFormats/unsanitized}}`, but some MIME types(like `image/png`) should be treated like `unsanitized data types` and doesn't have to be in `{{ClipboardUnsanitizedFormats/unsanitized}}`.

Currently Webkit and Chromium allow unsanitized `image/png` read/write by-default so web authors don't have to specify it in the new `{{ClipboardUnsanitizedFormats/unsanitized}}` option during `read()`.

index.bs
snianu: hm I guess we don't expose any MIME type that isn't supported by DataTransfer APIs or the default paste event, so I think we can remove it.

index.bs
```diff
@@ -542,6 +542,19 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 			* Custom format [=string/starts with=] `"web "`("web" followed by U+0020 SPACE) prefix
 				and suffix (after stripping out `"web "`) passes the [=parsing a MIME type=] check.
 
+	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
+
+		The implementation MAY recognize the native OS clipboard format description
+		for the following data types, to be able to populate the
+		{{ClipboardItem}} with the correct [=/MIME type=], and
+		set the correct data format on the OS clipboard in response to copy and cut
```
sanketj, line 550: Is it right to say in response to "copy and cut events" given that this doesn't apply to the `DataTransfer` or `execCommand` APIs?

snianu: Removed.

index.bs
snianu: Clarified few things.

index.bs
```diff
@@ -816,7 +808,11 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
-		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are in the [=unsanitized data types=].
+		<dfn>Optional unsanitized data types</dfn> are [=representation/mime type=]s that MUST not be sanitized by the user agent. The valid [=optional unsanitized data types=] are listed below:
```
sanketj, line 811: nit: You use "MUST not" here vs. "MUST NOT" above. Can we make these consistent?

sanketj, line 811: Can we move the definition of "optional unsanitized data types" up below the definition of "unsanitized data types"? That would be better for readability.

index.bs
```diff
@@ -833,7 +829,7 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 				1. For each |format| in |formats|["{{ClipboardUnsanitizedFormats/unsanitized}}"]:
 
-					1. If |format| is not in [=unsanitized data types=], then [=reject=] |p| with |format| {{"NotAllowedError"}} {{DOMException}} in |realm|.
+					1. If |format| is not in [=optional unsanitized data types=], then [=reject=] |p| with |format| {{"NotAllowedError"}} {{DOMException}} in |realm|.
```
sanketj, line 832: What's the reason this needs to be done in a separate for each loop here vs. inside the loop on line 866?

sanketj (changes requested):

index.bs
```diff
@@ -544,17 +544,9 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
-		set the correct data format on the OS clipboard in response to copy and cut
-		events.
-
-			These data types must not be sanitized and should be exposed by UAs
-			if a corresponding native type exists on the clipboard:
-
-			* text/html
+			These data types MUST NOT be sanitized by UAs:
 
+			* image/png (which should remain unsanitized to preserve meta data)
```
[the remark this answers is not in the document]
snianu: Done.

index.bs
snianu: In the Chromium's implementation we don't post the async task to read the data from the system clipboard for formats that aren't in `unsanitized` list if this check fails when `read()` is called. That makes sense as reading those formats are waste of resources if we end up throwing an exception anyway. This is why I had to move it outside the step that posts parallel tasks to do work.

index.bs
snianu: Done.

snianu: Done.

index.bs
```diff
@@ -542,6 +542,15 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 				and suffix (after stripping out `"web "`) passes the [=parsing a MIME type=] check.
 
+	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
+
+			These data types MUST NOT be sanitized by UAs:
+
+			* image/png
+
+			<p class=note>
+				The image/png MIME type is not sanitized to preserve meta data.
```
sanketj, line 552: This comment still doesn't really make it clear to me why we don't sanitize images. The statement "preserve meta data" is far too high level. Can we add more details here, or even better, an example?

index.bs
```diff
@@ -798,17 +812,28 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 		{{Clipboard/read()}} returns a {{Promise}} to [=clipboard items=] object that represents contents of [=system clipboard data=].
 		</p>
 
+		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that are [=optional unsanitized data types=].
```
sanketj, line 815:
```suggestion
		{{ClipboardUnsanitizedFormats/unsanitized}} is a [=sequence=] of {{DOMString}}s corresponding to the [=representation/mime type=] that the author wants to be treated as [=optional unsanitized data types=].
```

index.bs
sanketj: Ok, this sounds fine. It looks like we can bypass things like checking for permissions.

index.bs
```diff
@@ -837,13 +862,27 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+								1. If |format| is equal to [=representation/MIME type=], set |isUnsanitized| to true.
+
 						1. Set |representation|'s [=representation/data=] to |systemClipboardRepresentation|'s [=system clipboard representation/data=].
 
 							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
 						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data.
+						1. The user agent, MUST NOT sanitize |representation|'s [=representation/data=], if it satisfies the below conditions:
+						
+							1. |representation|'s [=representation/MIME type=] is in [=unsanitized data types=] list.
```
sanketj, line 879: Don't we also need to be include "optional unsanitized data types" here? The UA must not sanitize those either.

index.bs
```diff
@@ -542,6 +542,15 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 			* Custom format [=string/starts with=] `"web "`("web" followed by U+0020 SPACE) prefix
 				and suffix (after stripping out `"web "`) passes the [=parsing a MIME type=] check.
 
+	<h3 id="unsanitized-data-types-x"><dfn>Unsanitized data types</dfn></h3>
+
+			These data types MUST NOT be sanitized by UAs:
+
+			* image/png
```
sanketj, line 549: I wonder if we should include "optional unsanitized data types" in the set of data types in "unsanitized data types"? To my comment in the algorithm, this would make it clearer that UAs must treat optional and non-optional unsanitized data types the same after the optional unsanitized data types have been validated.

index.bs
sanketj: Don't think this is actually done. My ask was actually to move the definition of "optional unsanitized data types" right next to the definition of "unsanitized data types" (above or below is fine). That way, both definitions are in the same place and easier to process for the reader. Sound reasonable?

sanketj (changes requested):

index.bs
snianu: This was added as part of shipping the `image/png` MIME type in Chromium. I'm not very familiar with what metadata was getting stripped out due to sanitization, so I'll have to talk to the owner to understand all the details.

index.bs
snianu: Ack. Added per the above comment.

index.bs
snianu: Done.

index.bs
sanketj: In this case, I would just leave out the note entirely and just leave it as "image/png". It would still be worth understanding this aspect better and then adding some text in the spec as explanation as a follow up.

index.bs
snianu: Ack.

snianu: Done. PTAL.

index.bs
snianu: Done.

sanketj (approved):

snianu:
@evanstade I think the latest iteration addresses all your concerns, but please let me know if there is anything missing. I would like to restart the I2S process as I'm not receiving any response from Mozilla folks.

snianu:
Landing this PR. Please let me know if you have any comments and I'll address it in separate PRs.

snianu:
@zcorpan Sorry I created a new bug in Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1867331
Here is the request for position: https://chromestatus.com/feature/5192271976988672?gate=5118722541092864
For Webkit they [weren't interested](https://github.com/w3c/clipboard-apis/issues/150#issuecomment-974236367) in this feature as their sanitization process in async clipboard APIs aligns with the DataTransfer APIs.

saschanaz:
"Implementation commitment" doesn't sound like just filing bugs but more about implementation interest 🤔

saschanaz:
That's nice, but my understanding of commitment is that there at least should be some interest. Having a bug does not mean an interest, right?

saschanaz:
Sure. What I feel off here is the checkbox for Gecko is filled, when there's no public implementation commitment from Gecko.

snianu:
@saschanaz We can remove it from the spec if we only have one implementor interest, but waiting indefinitely for a signal is probably not the right way to decide whether there is interest from a particular browser. If the official stance of Firefox is negative, then please let me know and I can revert this PR and create another PR and mark it as BLOCKED. That way if Firefox or Webkit changes their stance and decide to support this feature in the future, we can unblock the PR and merge it.

saschanaz:
I see your point, but then how are we going to track which part doesn't have implementer consensus and which part does? Having an open PR is easier for that and doing so automatically prevents any part of the spec without consensus accidentally becomes CR.

snianu:
Note that the IDL and the language in the algorithms already allow non-supporting implementors to be in compliant with the spec while also allowing implementors who are interested in the new feature. I don't see any parts in the PR that is contentious unless I'm missing something?

saschanaz:
I'm talking about the general process rather than the PR itself. And also I don't think we'll want to add any kind of optional features just because it's okay to not implement it.

snianu:
> And also I don't think we'll want to add any kind of optional features without consensus

We [spent over two years trying to get consensus](https://github.com/w3c/clipboard-apis/issues/150) on this API, but we couldn't. Safari has [clearly stated their position](https://github.com/w3c/clipboard-apis/issues/150#issuecomment-974594001), so since you commented that you are not in favor of an optional parameter, should I mark Firefox's position as negative?

saschanaz:
Our position will be announced in our standards-position issue probably by Edgar. I definitely see your frustration, but still I think merging PRs without consensus makes it hard to make sure every part of the spec has consensus (which is needed for recommendation track), and I see no benefits to merge it early.

snianu:
@saschanaz Let's discuss this in the EditingWG meeting which is going to happen next Thursday. Without any clear stance on this API, it's hard to assume whether we have consensus or not. If you can officially state your position before that, we can definitely revert the PR, else, we will have to decide what to do with it in the EditingWG meeting where hopefully someone from Firefox can give us clear official stance on this issue. Thanks!

PR #214, opened 2024-05-01
The permission request stuff is no longer in contention

martinthomson:
I think.

I also think that a great deal more needs to be done with this explainer to get up to speed, so this change might not be appropriate at this time.

For instance, I think that the clipboardchange event is no longer really a thing (which would be good).

Implementation commitment:


annevk:
It's not in contention in so much that we decided to not do it, right? But if you remove that here as proposed the previous paragraph is now a bit unclear as how would you explicitly request it?

PR #215, opened 2024-05-22
Split implementor interest and implementation bugs from PR template

### LINKED ISSUE #207: Merge a PR only with explicit implementor consensus (followup)
See the discussion in https://github.com/w3c/clipboard-apis/pull/197#issuecomment-1841380176. It has Agenda+ but in a closed PR, so I'm opening a new issue to make things a bit more clear.

saschanaz:
Closes #207

I did not completely replace the existing template with the WHATWG one, although I would be okay with that.

Once we do this, we should make sure other Editing WG repositories get the same templates.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,10 +1,12 @@
 Closes #????
 
 For normative changes, the following tasks have been completed:
-
+ * [ ] At least two implementers are interested (and none opposed):
+   * …
+   * …
  * [ ] Modified Web platform tests (link to pull request)
 
-Implementation commitment:
+Implementation bugs are filed:
 
  * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
  * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,10 +1,12 @@
 Closes #????
 
 For normative changes, the following tasks have been completed:
-
+ * [ ] At least two implementers are interested (and none opposed):
```
snianu, line 4: > (and none opposed)

Any reason why this is an "and"?

snianu: Also tagging few more people to comment on this. @dandclark @sanketj

johanneswilm: This is fine for specs that have largely beenimplemented and where only minor things are to be changed added over time (like this one).
But the charter of our working groups has quite a different focus, opening for experimental ideas and new suggestions on how things could work that are being worked on collaboratively over time without there being a consensus among browser makets that this is actually tbe approach one ends up choosing.
In conclusion: this template could work for tjis tepository and sone others if the maintainers of those documents agree with it, but it should automatically be applied to all our repos.

sanketj: I agree that a template like this would be helpful, but I'm not sure if we should use "At least two implementers are interested (and none opposed)" as the condition for being able to land normative spec changes. This is a space where browsers already do not align with one another on several parts of the spec. Examples include: clipboard permission policy, web custom formats, sanitization. In the absence of interop, it has worked well in the past to at least document the behaviors of each engine in the spec. We should absolutely strive for interop with every change we make, but perhaps we need a better condition that allows for cases where browsers may have by-design differences.

Thoughts?

sanketj (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
annevk: If engines cannot come to agreement, that is what implementation-defined is for. Which if used to the extent you suggest is more indicative of failure of the standardization process than anything else.

A specification shouldn't discuss concrete implementations. That's not what it's for.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,10 +1,12 @@
...
+   * …
+   * …
  * [ ] Modified Web platform tests (link to pull request)
 
-Implementation commitment:
+Implementation bugs are filed:
 
  * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
```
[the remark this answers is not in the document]
annevk: Gecko is listed.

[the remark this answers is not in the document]
sanketj: Ah, it's below and was hidden out of view. Sorry I missed that.

.github/PULL_REQUEST_TEMPLATE.md
sanketj: Could you share an example of how you propose "implementation-defined" behaviors should be documented?

The reality of this space is that browsers have by-design differences that have prevented interop from being achieved. In the absence of interop, documenting behaviors in the spec can still be useful to minimize compat issues that could arise from a lack of interop.

johanneswilm:
Having a template like this or a similar one seems like a good idea to the extent that it ensures that browser bugs are filed correctly. But it should not change what we are working or how we are working on it in the WG.

.github/PULL_REQUEST_TEMPLATE.md
annevk: We don't do that anywhere in the web platform. I'm not sure why we'd start here.

johanneswilm: This has been done over the past 12-15 years, starting with changes observed in execCommand ( https://w3c.github.io/editing/docs/execCommand/ ) that were found that couldn't easily be removed.

annevk: There's a number of notes there that I don't think make sense to publish as part of a standard, but I don't think there's anything normative there that does this.

johanneswilm: @annevk That is the work of the previous editor and I agree that the wording doesn't follow what I would expect for this kind of document. Maybe the current editor, @zcorpan, will want to change that eventually. Nevertheless, that was the beginning of documenting differences between browsers.

I don't think we would want to document differences in a "normative" way. I guess that would mean that one specific browser HAS to work in a specific way that is different from another browser.

PR #219, opened 2024-08-01
Remove issues claiming "clipboard-write" was removed

foolip:
https://github.com/w3c/clipboard-apis/pull/164 is still open, and the permission is defined here: https://w3c.github.io/clipboard-apis/#clipboard-permissions

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1024,8 +1024,6 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 				1. Let |r| be the result of running [=check clipboard write permission=].
 
-					Issue: clipboard-write was removed in https://github.com/w3c/clipboard-apis/pull/164.
-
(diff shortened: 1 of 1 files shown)


saschanaz:
I would be happier if we just move this permission check to a separate doc for now. If not we should still have something like feature-at-risk.

foolip:
@saschanaz that seems like a discussion for https://github.com/w3c/clipboard-apis/pull/164 which is now stalled. In the meantime, this issue text is wrong. Updating it some other way to be accurate would be fine too, I just found it confusing when trying to understand what's going on here.

saschanaz:
Feature-at-risk would just explain the current situation IMO without much change in the spec, right?

annevk:
Making it

> Issue: clipboard-write does not have agreement. See ... for more details.

could work.

PR #220, opened 2024-08-08
Update PR template

sanketj:
Per resolution to w3c/editing#463, w3c/editcontext#100 introduced a PR template that was agreed upon by the Editing WG. This change introduces the same template in this repository.


saschanaz (approved):
(Just to log: I still think we should ultimately require explicit implementation interest, but this is certainly a progress)

PR #234, opened 2025-03-18
chore: fix markup errors for the automatic publication

siusin:
***
<a href="https://pr-preview.s3.amazonaws.com/siusin/clipboard-apis/pull/234.html" title="Last updated on Mar 18, 2025, 10:36 AM UTC (7087ea6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/234/2eebe17...siusin:7087ea6.html" title="Last updated on Mar 18, 2025, 10:36 AM UTC (7087ea6)">Diff</a>

PR #238, opened 2025-05-05
Fix `clipboard.readText()` algorithm

### LINKED ISSUE #236: Clarifying `clipboard.readText()` behaviour
I found the steps involved in creating a representation to be a bit unclear.

> 3. Let representation be a new [representation](https://w3c.github.io/clipboard-apis/#representation).
> 4. If representation’s [MIME type](https://w3c.github.io/clipboard-apis/#representation-mime-type) [essence](https://mimesniff.spec.whatwg.org/#mime-type-essence) is "text/plain", then:

How can it be "text/plain"? Shouldn't we check the essence of `mimeType`, which is:

> 1. Let mimeType be the result of running the [well-known mime type from os specific format](https://w3c.github.io/clipboard-apis/#well-known-mime-type-from-os-specific-format) algorithm given systemClipboardRepresentation’s [name](https://w3c.github.io/clipboard-apis/#system-clipboard-representation-name).

Lastly what should we actually do with data from clipboard? Spec is not clear about it, currently Servo implementation resolve representation's [data](https://w3c.github.io/clipboard-apis/#representation-data) with it.

EdgarChen:
Fixes #236.

 * [ ] Editing WG resolution on the proposed changes, with at least two implementers participating and not objecting:
   * [ ] WebKit
   * [ ] Chromium
   * [ ] Gecko

 * [ ] For browsers that are shipping the feature, implementation bugs are filed for the proposed changes (link to bug, or write "Not Implementing"):

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -956,51 +956,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 					Issue: Add definition of sanitized copy.
 
-				1. [=Queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to perform the below steps:
+				1. For each |systemClipboardItem| in |data|:
 
-					1. For each |systemClipboardItem| in |data|:
+					1. For each |systemClipboardRepresentation| in |systemClipboardItem|:
 
-						1. For each |systemClipboardRepresentation| in |systemClipboardItem|:
-
-							1. Let |mimeType| be the result of running the [=well-known mime type from os specific format=] algorithm given |systemClipboardRepresentation|'s [=system clipboard representation/name=].
+						1. Let |mimeType| be the result of running the [=well-known mime type from os specific format=] algorithm given |systemClipboardRepresentation|'s [=system clipboard representation/name=].
 							
-							1. If |mimeType| is null, continue this loop.
+						1. If |mimeType| is null, continue this loop.
 
-							1. Let |representation| be a new [=representation=].
+						1. If |mimeType|'s [=MIME type/essence=] is "text/plain", then:
 
-							1. If |representation|'s [=representation/MIME type=] [=MIME type/essence=] is "text/plain", then:
+							1. Resolve |p| with |systemClipboardRepresentation|'s [=system clipboard representation/data=].
 
-								1. Set |representation|'s [=representation/MIME type=] to |mimeType|.
+							1. Abort these steps.
 
-								1. Let |representationDataPromise| be the |representation|'s [=representation/data=].
+				1. Resolve |p| with empty string.
 
-								1. [=promise/React=] to |representationDataPromise|:
-
-									1. If |representationDataPromise| was fulfilled with value |v|, then:
-
-										1. If |v| is a {{DOMString}}, then follow the below steps:
-
-											1. Resolve |p| with |v|.
-
-											1. Return |p|.
-
-										1. If |v| is a {{Blob}}, then follow the below steps:
-
-											1. Let |string| be the result of [=UTF-8 decoding=] |v|'s underlying byte sequence.
-
-											1. Resolve |p| with |string|.
-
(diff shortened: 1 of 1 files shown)


EdgarChen:
The markup error should be fixed by #234.

index.bs
```diff
@@ -956,51 +956,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 
 					Issue: Add definition of sanitized copy.
 
-				1. [=Queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to perform the below steps:
-
-					1. For each |systemClipboardItem| in |data|:
+				1. For each |systemClipboardItem| in |data|:
```
annevk, line 959: Existing issue: "For each ... of ..." (applies a few times)

index.bs
```diff
@@ -956,51 +956,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
-											1. Return |p|.
-
-										1. If |v| is a {{Blob}}, then follow the below steps:
-
-											1. Let |string| be the result of [=UTF-8 decoding=] |v|'s underlying byte sequence.
-
-											1. Resolve |p| with |string|.
-
-											1. Return |p|.
+						1. If |mimeType| is null, continue this loop.
```
annevk, line 965:
```suggestion
						1. If |mimeType| is null, then [=iteration/continue=].
```

index.bs
```diff
@@ -956,51 +956,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 
-									1. If |representationDataPromise| was rejected, then:
+						1. If |mimeType|'s [=MIME type/essence=] is "text/plain", then:
```
annevk, line 967:
```suggestion
						1. If |mimeType|'s [=MIME type/essence=] is "`text/plain`":
```

annevk (commented):
You removed task queueing, but you cannot resolve or reject promises while in parallel without queueing a task first.

EdgarChen:
> You removed task queueing, but you cannot resolve or reject promises while in parallel without queueing a task first.

Oh, right, I added it back. Thank you.

index.bs
```diff
@@ -958,49 +958,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
-
-									1. If |representationDataPromise| was rejected, then:
-
-										1. [=Reject=] |p| with {{"NotFoundError"}} {{DOMException}} in |realm|.
-
-										1. Return |p|.
+								1. Abort these steps.
 
-					1. [=Reject=] |p| with {{"NotFoundError"}} {{DOMException}} in |realm|.
+					1. Resolve |p| with empty string.
```
annevk, line 975:
```suggestion
					1. Resolve |p| with the empty string.
```

index.bs
```diff
@@ -958,49 +958,23 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 							1. Let |mimeType| be the result of running the [=well-known mime type from os specific format=] algorithm given |systemClipboardRepresentation|'s [=system clipboard representation/name=].
 							
-							1. If |mimeType| is null, continue this loop.
+							1. If |mimeType| is null, then [=iteration/continue=].
 
-							1. Let |representation| be a new [=representation=].
+							1. If |mimeType|'s [=MIME type/essence=] is "`text/plain`":
 
-							1. If |representation|'s [=representation/MIME type=] [=MIME type/essence=] is "text/plain", then:
+								1. Resolve |p| with |systemClipboardRepresentation|'s [=system clipboard representation/data=].
```
annevk, line 971: `[=system clipboard representation/data=]` can be a string or a blob, right? But we want to always return a string. How is that going to work now?

annevk (commented):
The commit message body could explain a bit more about why this change is being made.

PR #239, opened 2025-06-23
Add comprehensive clipboardchange event API specification

### LINKED ISSUE #230: Add privacy considerations for the clipboardchange event
The Asynchronous Clipboard API allows reading the clipboard (including copied passwords), at any time by default, but the privacy considerations for it suggest that UAs might restrict it to after a user gesture. If a page is trying to steal data, it has to guess when that data's likely to be on the clipboard. It can poll, but polling is potentially suspicious, and a UA might alert the user or take other countermeasures.

[`clipboardchange`](https://www.w3.org/TR/clipboard-apis/#clipboard-event-clipboardchange) changes the behavior to allowing the page to trivially track the entire history of the clipboard. This should be called out in the privacy considerations, and the WG should describe any potential mitigations there, since the first current mitigation in that section doesn't apply to this new event.

### LINKED ISSUE #232: Proposal: Identifying whether the clipboard has changed without re-reading it
Hi,  
I’d like to propose an addition to the Clipboard API \- a way to efficiently check whether the clipboard has changed between two points in time without actually reading it. The primary use case that I have in mind would be to optimize the remote desktop solutions’ heuristics for local clipboard change detection. Another potential use case could be within the web versions of office-like applications, as they usually integrate context menus with information on what currently exists in the clipboard \- the exact interest of their developers in this however remains to be determined.  
You can find a more detailed explainer at [https://github.com/explainers-by-googlers/clipboard-contents-id](https://github.com/explainers-by-googlers/clipboard-contents-id). Could we schedule a discussion at an upcoming meeting to figure out:

1. Is there consensus to address this problem?  
2. If not, what extra data could help show that the use case is important?  
3. If the use case does have consensus:  
    1\. Can we check the explainer into [https://github.com/w3c/clipboard-apis](https://github.com/w3c/clipboard-apis)?  
    2\. What concerns do people have about the proposed API shape?  
    3\. How should this interact with the proposed `clipboardchange` event?

roraja:
## Summary

This PR adds a complete specification for the `clipboardchange` event to the W3C Clipboard API specification, including the missing `ClipboardChangeEventInit` dictionary and enhanced documentation based on implementation experience and the Microsoft Edge explainer.

The bikeshed rendered version can be verified from here: 
https://rohanraja.github.io/cr-clipboard/clipboardchange/spec/#clipboard-event-clipboardchange

### Added ClipboardChangeEventInit Dictionary
- **New:** `ClipboardChangeEventInit` dictionary with `types` sequence parameter
- **Documentation:** Clear specification of available mandatory data types on system clipboard
- **Security:** Explicit exclusion of custom data types to prevent fingerprinting

### Enhanced ClipboardChangeEvent Interface
- **Improved:** Detailed documentation for the `types` attribute behavior
- **Clarified:** Event firing conditions and focus requirements
- **Added:** Constructor steps algorithm following W3C specification patterns

## Implementation Status

- ✅ **Chromium**: Tracking bug [#41442253](https://issues.chromium.org/issues/41442253)
- ✅ **Chrome Platform Status**: [#5085102657503232](https://chromestatus.com/feature/5085102657503232)
- 🔄 **TAG Review**: [w3ctag/design-reviews#1017](https://github.com/w3ctag/design-reviews/issues/1017)

## Related Documentation

- **Microsoft Edge Explainer**: [clipboard-change-event-explainer.md](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ClipboardAPI/ClipboardChangeEvent/clipboard-change-event-explainer.md)
- **Intent to Prototype**: Posted November 2024
- **Feature Name**: ClipboardChangeEvent

## Testing

The specification includes comprehensive examples that can be used for:
- Web Platform Tests (WPTs) development
- Cross-browser compatibility validation
- Security and privacy testing scenarios

## Security & Privacy Considerations

- **No permissions required**: Event only exposes data types, not contents
- **Fingerprinting mitigation**: Custom data types excluded from event payload
- **Focus-gated**: Events only fire when document has focus
- **Privacy-preserving**: No clipboard contents exposed in event

---

**Contacts**: roraja@microsoft.com, ragoulik@microsoft.com

Fixes #230 

 * [X] Editing WG resolution on the proposed changes, with at least two implementers participating and not objecting:
   * [X] WebKit
   * [X] Chromium
   * [X] Gecko

***
<a href="https://pr-preview.s3.amazonaws.com/roraja/clipboard-apis/pull/239.html" title="Last updated on Nov 14, 2025, 2:42 AM UTC (7fdd833)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/239/377cd12...roraja:7fdd833.html" title="Last updated on Nov 14, 2025, 2:42 AM UTC (7fdd833)">Diff</a>


index.bs
```diff
@@ -224,12 +224,86 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
 			* Scripts that use the [[#async-clipboard-api]] to write to the clipboard
 			* Actions that update the clipboard outside the user agent
 
-			If the clipboard contents are changed outside the user agent, then the
-			{{GlobalEventHandlers/clipboardchange}} event MUST fire when the user agent regains focus.
+			If the clipboard contents are changed while the document is not in focus, a single
+			{{GlobalEventHandlers/clipboardchange}} event MUST fire when the document regains focus,
+			even if the clipboard changed multiple times.
+			No event fires when the clipboard changes while the document is out of focus, which helps prevent malicious applications from monitoring the clipboard in the background.
```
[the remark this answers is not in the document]
roraja: Fixed, thanks.

index.bs
```diff
@@ -224,12 +224,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+			<div class="algorithm" data-algorithm="fire-clipboardchange-event">
+			<h5 id="h-fire-clipboardchange-event">To fire a clipboardchange event</h5>
+
+			To <dfn export>fire a clipboardchange event</dfn> given a {{Document}} <var>document</var>:
+
+			1. If <var>document</var> does not have [=sticky activation=], then return.
+
+			2. Let <var>global</var> be <var>document</var>'s [=relevant global object=].
+
+			3. If <var>document</var> has <a>focus</a>:
```
zgroza, line 236: Can maybe [system focus](https://html.spec.whatwg.org/multipage/interaction.html#system-focus) be used here and in other places? The current specification mentions only [has focus](https://html.spec.whatwg.org/multipage/interaction.html#has-focus-steps) in the [Privacy and the Asychronous Clipboard API](https://w3c.github.io/clipboard-apis/#privacy-async) section, which directly references this term.
```suggestion
			3. If <var>document</var> has [=system focus=]:
```

Alternatively, just like the spec currently does,
```suggestion
			3. If <var>document</var> [=has focus steps|has focus=]:
```

index.bs
```diff
@@ -224,12 +224,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+				1. Let <var>types</var> be a [=list=] of [=mandatory data types=] available on the [=system clipboard=].
+				2. Let <var>changeId</var> be a unique identifier for this clipboard change, generated by applying an origin-specific transformation to a system-level clipboard change identifier, ensuring that different origins receive different identifiers for the same clipboard change while maintaining consistency within the same origin.
```
zgroza, line 238: Maybe let's use partitions here? [Specification](https://w3ctag.github.io/privacy-principles/#user-agent-recognition) says that
> [User agents](https://infra.spec.whatwg.org/#user-agent) should prevent people from being [recognized](https://w3ctag.github.io/privacy-principles/#dfn-recognize) across [partitions](https://w3ctag.github.io/privacy-principles/#dfn-partition) unless they intend to be recognized.

```suggestion
				2. Let <var>changeId</var> be a unique identifier for this clipboard change, generated by applying an origin-specific transformation to a system-level clipboard change identifier, ensuring that different [=dfn-partition|partitions=] receive different identifiers for the same clipboard change while maintaining consistency within the same [=dfn-partition|partition=].
```

index.bs
```diff
@@ -224,12 +224,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+				sequence&lt;DOMString&gt; types = [];
+				DOMString changeId = "";
+			};
+			</pre>
+
+			: <dfn for="ClipboardChangeEventInit" data-export>types</dfn>
+			:: A sequence of {{DOMString}} representing the [=mandatory data types=] available on the [=system clipboard=].
+
+			: <dfn for="ClipboardChangeEventInit" data-export>changeId</dfn>
+			:: A {{DOMString}} representing a unique identifier for this clipboard change operation.
```
zgroza, line 286: I wonder a bit whether a random `BigInt` wouldn't be more straightforward here. Having a UUID in the form of a `DOMString` here forces:
* user agent to convert to/from the standard UUID text serialization (and worry about adhering to particular format)—which requires either some library dependency or a custom implementation
* web applications to compare strings (or convert them), which might be *a bit* less straightforward than just comparing integers

Those are arguably very small overheads, so it's not really a very strong preference—it might however improve sheer simplicity.

index.bs
```diff
@@ -224,12 +224,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
 			* Actions that update the clipboard outside the user agent
 
-			If the clipboard contents are changed outside the user agent, then the
-			{{GlobalEventHandlers/clipboardchange}} event MUST fire when the user agent regains focus.
+			<div class="algorithm" data-algorithm="fire-clipboardchange-event">
+			<h5 id="h-fire-clipboardchange-event">To fire a clipboardchange event</h5>
+
+			To <dfn export>fire a clipboardchange event</dfn> given a {{Document}} <var>document</var>:
+
+			1. If <var>document</var> does not have [=sticky activation=], then return.
```
zgroza, line 232: This troubles me a bit, as it doesn't work well with the current spec and use cases. In case of the persistent clipboard permissions, site could theoretically read from clipboard on focus—but before gaining sticky activation—and then not be notified when the read data is already stale.
Maybe it would be better to either:
* Move this like focus to the current step `3.`—and set the pending flag also if the sticky activation is not yet there?
* Add sticky activation requirement to the [Privacy and the Asychronous Clipboard API](https://w3c.github.io/clipboard-apis/#privacy-async) of the entire API overall? This could potentially break current websites—on the other hand, would align with the current Firefox and Safari implementation, which requires transient user activation (which is impossible to get without sticky activation) anyway.
* Add a provision that if the persistent clipboard permission is granted (provided that the user agent supports it), sticky activation check is skipped? It doesn't make much sense then anyway, as the application can just read the clipboard—which provides much more sensitive data than just a change notification.

index.bs
roraja: Since the changeID is mostly locally unique and not something that we would need to store in a database or persist, a BigInt indeed might be good enough for us. Though there are less chances of collisions in UUID, still for our use-case this should not be a problem. 

Though I couldn't find any existing spec where BigInt is used as a local identifier. I will still go ahead and change this ID's type to BigInt to prefer simplicity as you mentioned.

index.bs
```diff
@@ -224,12 +227,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+
+			To <dfn export>fire a clipboardchange event</dfn> given a {{Document}} <var>document</var>:
+
+			1. If <var>document</var> does not have [=sticky activation=] and <var>document</var> does not have permission to read from the clipboard, then return.
+
+			2. Let <var>global</var> be <var>document</var>'s [=relevant global object=].
+
+			3. If <var>document</var> has [=system focus=]:
+				1. Let <var>types</var> be a list of [=mandatory data types=] available on the [=system clipboard=].
+				2. Let <var>changeId</var> be a unique identifier for this clipboard change, generated by applying a partition-specific transformation to a system-level clipboard change identifier, ensuring that different [=partitions=] receive different identifiers for the same clipboard change while maintaining consistency within the same [=partition=].
```
annevk, line 241: You need to actually obtain the partition somehow. Maybe this should repurpose the storage key?

index.bs
```diff
@@ -224,12 +227,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+				3. Let <var>changeId</var> be a unique identifier for this clipboard change, generated by applying a partition-specific transformation to a system-level clipboard change identifier, ensuring that different [=partitions=] receive different identifiers for the same clipboard change while maintaining consistency within the same [=partition=].
+				4. Let <var>eventInit</var> be a new {{ClipboardChangeEventInit}} dictionary with its {{ClipboardChangeEventInit/types}} member set to <var>types</var> and its {{ClipboardChangeEventInit/changeId}} member set to <var>changeId</var>.
+				5. Let <var>global</var> be <var>document</var>'s [=relevant global object=].
+				6. [=Fire an event=] named {{GlobalEventHandlers/clipboardchange}} at <var>global</var>, using {{ClipboardChangeEvent}}, with <var>eventInit</var>.
+
+			</div>
+
+			For documents in nested browsing contexts, the {{GlobalEventHandlers/clipboardchange}} event fires independently in each {{Document}} based on its own focus state. A clipboard change will fire events in all focused documents across the browsing context tree that have [=sticky activation=] or persistent clipboard permissions.
```
annevk, line 264: Only a single document can have focus, no?

roraja: Fixed, thanks

index.bs
```diff
@@ -224,12 +224,166 @@ url: https://w3c.github.io/permissions/#permissions-task-source; type: dfn;
...
+			1. If <var>document</var> does not have [=sticky activation=], then return.
+
+			2. Let <var>global</var> be <var>document</var>'s [=relevant global object=].
+
+			3. If <var>document</var> has <a>focus</a>:
+				1. Let <var>types</var> be a [=list=] of [=mandatory data types=] available on the [=system clipboard=].
+				2. Let <var>changeId</var> be a unique identifier for this clipboard change, generated by applying an origin-specific transformation to a system-level clipboard change identifier, ensuring that different origins receive different identifiers for the same clipboard change while maintaining consistency within the same origin.
+				3. Let <var>eventInit</var> be a new {{ClipboardChangeEventInit}} dictionary with its {{ClipboardChangeEventInit/types}} member set to <var>types</var> and its {{ClipboardChangeEventInit/changeId}} member set to <var>changeId</var>.
+				4. [=Fire an event=] named {{GlobalEventHandlers/clipboardchange}} at <var>global</var>, using {{ClipboardChangeEvent}}, with <var>eventInit</var>.
+			4. Otherwise:
```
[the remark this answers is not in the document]
roraja: Fixed, thanks

index.bs
annevk: The Storage Standard doesn't expose "storage shelf" for external usage so that would require new specification API of sorts.

roraja: I have updated the spec as per this discussion keeping the implementation detail non-normative while keeping the expectation from changeID normative. Would this work ?

johanneswilm (approved):

NAME-sovsec:
https://github.com/w3c/clipboard-apis/commit/efdccf07a0cc518b4faba4095a07b0a2e0f06c83.diff

johanneswilm:
TPAC 2025:
> Resolution is to include the feedback from Anne
And adopt the resolution

Issue #240, opened 2025-08-04
Selective Clipboard Format Read


shwetabin:
Hi,
I'd like to propose an addition to the Clipboard API - a way to efficiently get specific MIME types in the Async Clipboard Read API. The primary use case I have in mind involves web applications that support rich content editing, such as document editors, email clients, and data grids, which routinely deal with multiple types of clipboard payloads, including large HTML fragments, images, and custom MIME types. These applications often implement features like "Paste as plain text" or "Paste with formatting," where only a subset of the clipboard data is needed. You can find more detailed explainer at https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ClipboardAPI/SelectiveClipboardFormatRead/explainer.md. Could we schedule a discussion at an upcoming meeting to figure out:

1. Is there consensus to address this problem?
2. If not, what extra data could help show that the use case is important?
3. If the use case does have consensus:
    1. Can we check the explainer into  https://github.com/w3c/clipboard-apis?
    2. What concerns do people have about the proposed API shape?

annevk:
Thank you @EdgarChen for pointing that out! I learned something new there about Safari's implementation that I had not realized and was seemingly missed by all of us during the last meeting. That `getType()` is already lazy and needs to be called before the clipboard changes. And it's lazy in a way that I think works.

That is, if you do `read()`, followed by the user doing a fresh copy, followed by `getType()`, `getType()` will throw/return something empty.

That model seems vastly preferable over having to cache the entire clipboard item in memory/some side table store each time `read()` is called. Especially with the proposed `clipboardchange` event making it very clear at which point invalidation for a prior `read()` happens.

johanneswilm:
TPAC 2025:
> Rohan
I hope everyone is aware of this proposal to optimize MIME type reading 
We want to optimize where the web author can request specific MIME types
We have two options, one was to specifically ask for a set of MIME types
The second option can cause certain backward compatibility challenges
This changes the behaviour of the clipboard item as well
With this change, this will give you the latest data every time it is called
We were concerned that some website were replying on the clipboard 
To have a persistent state 
Say a clipboard timeline application
So there was a backward compatibility list
To move forward we have added this in chromium
There are a few more concern around this
The backward compatibility issue
The user permission part
So if we are moving the read to the get type, do we want to move the
Permission to getType. 
The other scenario is that you do the permission first. 
Since the permission is refused.
Smaug
The permission model is where this is difficult
They were not expecting and error from getType
[...]

PR #248, opened 2026-03-26
Defer OS clipboard read from clipboard.read() to ClipboardItem.getType()

shwetabin:
Spec change to defer OS clipboard reads from `clipboard.read()` to
  `ClipboardItem.getType()`.

  Previously, `clipboard.read()` eagerly fetched all clipboard payload bytes from
  the OS at call time. This change specifies that `clipboard.read()` enumerates
  only the available format names (metadata), and the actual OS data read is
  deferred to when the author calls `getType()` on the resulting `ClipboardItem`.

  Specific changes:
  - `clipboard.read()`: initializes each representation's data to a new unresolved
    promise instead of eagerly resolving it with OS clipboard bytes. Resolves the
    existing Issue: "It should be possible to read the data asynchronously from
    the system clipboard after the author calls getType".
  - `ClipboardItem.getType()`: adds a deferred OS read block that runs in parallel,
    reads bytes via the `os specific well-known format` algorithm, applies
    sanitization (same rules as before), and queues resolution on the clipboard
    task source per Infra conventions.
 w3c issue line - https://github.com/w3c/clipboard-apis/issues/240

***
<a href="https://pr-preview.s3.amazonaws.com/shwetabin/clipboard-apis/pull/248.html" title="Last updated on Jun 23, 2026, 8:40 AM UTC (e5cee33)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/clipboard-apis/248/ab5f530...shwetabin:e5cee33.html" title="Last updated on Jun 23, 2026, 8:40 AM UTC (e5cee33)">Diff</a>


annevk:
Can you mention in the commit message that this fixes #240? At least I assume that's the goal.

siusin:
hi @shwetabin , could you link your [GitHub account with your W3C account](https://www.w3.org/users/myprofile/connectedaccounts) please? This will help our IPR bot recognize you as a working group member.

index.bs
```diff
@@ -918,6 +928,44 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
+
+							1. Let |webCustomFormatMap| be the [=os specific custom map name=].
+
+							1. Read |webCustomFormatMap| from the [=system clipboard=].
+
+							1. Let |webCustomFormatMapString| be the JSON string deserialized from |webCustomFormatMap|.
+
+							1. Let |osFormatName| be the value in |webCustomFormatMapString| whose key matches |mimeType| [=serialize a MIME type|serialized=].
+
+							1. If |osFormatName| is not found, then [=queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to [=reject=] |p| with a {{"NotFoundError"}} {{DOMException}} in |realm|, then abort these steps.
```
EdgarChen: We reject the `p` which is new created Promise for `read()`, and then later we still `React to representationDataPromise`, is this intentional? Or when does the `representationDataPromise` be fulfilled or rejected? Or maybe I miss something.

index.bs
```diff
@@ -918,6 +928,44 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
 
 				1. If |representation|'s [=representation/MIME type=] is |mimeType| and |representation|'s [=representation/isCustom=] is |isCustom|, then:
 
+					1. If [=this=]'s [=ClipboardItem/clipboard change count at read=] is not null, and the current [=clipboard change count=] is not equal to [=this=]'s [=ClipboardItem/clipboard change count at read=], then [=reject=] |p| with a {{"DataError"}} {{DOMException}} in |realm|, and return |p|.
```
EdgarChen: So if the `clipboard change count` is null, the ClipboardItem is created from the script? Maybe we can have some note somewhere.

EdgarChen (changes requested):

index.bs
shwetabin: Yes, exactly — a null clipboard change count at read means the ClipboardItem was constructed by the author (e.g., via the constructor), not obtained from read(). Added a clarifying note at the check. See latest commit.

index.bs
shwetabin: To clarify, |p| is the promise created by getType() (step 5), not by read(). The return |p| in the stale-data check exits the getType() method entirely, so the React to representationDataPromise steps are never reached in that case.

index.bs
```diff
@@ -918,6 +928,46 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
+
+						Note: This ensures that stale data is never returned. If the [=system clipboard=] contents have changed since {{Clipboard/read()}} was called, this MUST fail rather than returning data that does not correspond to the current [=system clipboard=] state.
+
+						Note: A null [=ClipboardItem/clipboard change count at read=] indicates that this {{ClipboardItem}} was constructed directly by the author (e.g., via {{ClipboardItem/ClipboardItem(items, options)}}), not obtained from {{Clipboard/read()}}. In that case, no stale-data check is needed because the data does not originate from the [=system clipboard=].
+
+					1. If |representation|'s [=representation/data=] is not yet resolved, run the following steps [=in parallel=]:
```
EdgarChen: We check whether the `representation`'s `data` (which is [promise](https://www.w3.org/TR/clipboard-apis/#representation-data)) is already resolved. What if two `getType()` called twice in a row with the same MIME type? Would it be possible both `getType()` calls to try to resolve or reject the data promise?
Maybe we need some way to handle that case?

index.bs
```diff
@@ -918,6 +928,46 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
+
+							1. Let |webCustomFormatMapString| be the JSON string deserialized from |webCustomFormatMap|.
+
+							1. Let |osFormatName| be the value in |webCustomFormatMapString| whose key matches |mimeType| [=serialize a MIME type|serialized=].
+
+							1. If |osFormatName| is not found, then [=queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to [=reject=] |p| with a {{"NotFoundError"}} {{DOMException}} in |realm|, then abort these steps.
+
+						1. Else, let |osFormatName| be the result of running [=os specific well-known format=] given |mimeType|.
+
+						1. Let |rawData| be the result of reading bytes for |osFormatName| from the [=system clipboard=].
```
EdgarChen: Not sure how does the "the result of reading bytes .." work and I assume the result could be a failure or byte sequence. Is the data from `system clipboard representation` in https://w3c.github.io/clipboard-apis/#model or maybe something else?

shwetabin: Thanks for the feedback. I simplified this to avoid introducing a separate helper algorithm.
I removed the standalone “read bytes” definition and inlined the system clipboard data fetch directly in getType, so the spec now does the lookup at the call site (get current system clipboard item, find representation by osFormatName, then use its data).

index.bs
shwetabin: Yes, I covered this in the latest update.
We now cache the first same-type getType() Promise in the resolver map and return it for subsequent calls, so no duplicate read/settle path can occur.

index.bs
```diff
@@ -918,6 +934,56 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
+
+								1. Let |webCustomFormatMapString| be the JSON string deserialized from |webCustomFormatMap|.
+
+								1. Let |osFormatName| be the value in |webCustomFormatMapString| whose key matches |mimeType| [=serialize a MIME type|serialized=].
+
+								1. If |osFormatName| is not found, then [=queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to [=reject=] |p| with a {{"NotFoundError"}} {{DOMException}} in |realm|, then abort these steps.
+
+							1. Else, let |osFormatName| be the result of running [=os specific well-known format=] given |mimeType|.
+
+						1. Let |clipboardItem| be the [=system clipboard item=] currently held in the [=system clipboard=]. If there is no such item, then [=queue a global task=] on the [=clipboard task source=], given |realm|'s [=realm/global object=], to [=reject=] |p| with a {{"NotFoundError"}} {{DOMException}} in |realm|, then abort these steps.
```
annevk: Shouldn't this still be done while in parallel?

annevk: Also, which item are we talking about here? How are they mapped?

index.bs
```diff
@@ -898,6 +912,8 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
 
 			1. Let |realm| be [=this=]'s [=relevant realm=].
 
+			1. Let |originalType| be |type|.
```
annevk: Is |type| here completely not normalized? How can that reasonably be used as a map key?

index.bs
shwetabin: The mapping is implicit via the [=ClipboardItem/clipboard change count at read=] check earlier in this algorithm: if the [=clipboard change count=] hasn't changed since {{Clipboard/read()}}, the "currently held" [=system clipboard item=] is by construction the same one {{Clipboard/read()}} observed. In practice the [=system clipboard=] holds at most one item at a time, so "currently held" is unambiguous.

index.bs
shwetabin: Good catch — switched the map key from the raw |type| to a normalized one (serialize |mimeType|, prefixed with "web " when |isCustom| is true), so differently-cased spellings of the same type now collapse to a single entry.

index.bs
shwetabin: You're right — the steps from "Let |clipboardItem|…" through the blob construction had drifted out of the in parallel block; re-indented them so only the final queue a global task to resolve/reject crosses back to the event loop.

index.bs
annevk: I think we probably only want to use _mimeType_'s essence here, not the serialization including all parameters.

index.bs
annevk: Is that true on all platforms? But if that's the case we should make that clear in the data model. Because if in practice and in theory differ, it's certainly ambiguous.

index.bs
```diff
@@ -918,6 +934,60 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
 
 				1. If |representation|'s [=representation/MIME type=] is |mimeType| and |representation|'s [=representation/isCustom=] is |isCustom|, then:
 
+					1. If [=this=]'s [=ClipboardItem/clipboard change count at read=] is not null, and the current [=clipboard change count=] is not equal to [=this=]'s [=ClipboardItem/clipboard change count at read=], then [=reject=] |p| with a {{"DataError"}} {{DOMException}} in |realm|, and return |p|.
```
dandclark: I'm not totally sold on `DataError` being the right choice to reject. Does this come form WebKit's current behavior?
Worth considering some others that might be a better fit from https://webidl.spec.whatwg.org/#idl-DOMException-error-names:
- `InvalidStateError`
- `NotAllowedError`
- `SecurityError`

A comparison of the other things that can throw these errors and `DataError` might help grant some confidence of which might be the best fit.

index.bs
shwetabin: I checked it, and iOS supports multiple clipboard items. In the ClipboardItem definition, we added text stating that each clipboard item is linked to a corresponding system clipboard item, and getType() calls will read data from that system clipboard item.

index.bs
shwetabin: Good catch—InvalidStateError is the best match for our use case. I've updated the PR.

johanneswilm:
Call 2026-05-14:
> Clipboard - Keyboard Selection Format (PR 248): Rakesh Goulikar 
requested a status check on pull request 248 regarding keyboard 
selection format. Shweta Bindal stated that the changes are ready and 
welcomed further review from Mozilla or Safari. The team agreed to 
coordinate with Anne from Safari to ensure the changes are acceptable, 
particularly since Chromium plans to ship this feature imminently.

shwetabin:
Hi @EdgarChen and @annevk,
Please let me know if you have any further comments. Otherwise, I’ll go ahead and merge the PR.

index.bs
```diff
@@ -830,6 +832,20 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
 
 		A {{ClipboardItem}} object has an associated <dfn for="ClipboardItem">types array</dfn>, which is a {{FrozenArray}}&lt;{{DOMString}}&gt;.
 
+		A {{ClipboardItem}} object has an associated <dfn for="ClipboardItem">clipboard change count at read</dfn>, which is initially null. When the {{ClipboardItem}} is created from {{Clipboard/read()}}, this is set to the [=clipboard change count=] at the time the [=system clipboard data=] was snapshotted.
```
annevk: We have "When the {{ClipboardItem}} is created from {{Clipboard/read()}}" multiple times here, but this should probably be in the longer note below or removed completely as this is not normative. The normative requirements are in the `read()` algorithm after all.

annevk:
@shwetabin I think you should wait for @EdgarChen to review again.

index.bs
```diff
@@ -918,6 +934,60 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
+
+					1. If [=this=]'s [=ClipboardItem/clipboard change count at read=] is not null, then:
+
+						Note: This branch covers {{ClipboardItem}}s obtained from {{Clipboard/read()}}, for which the OS clipboard read is deferred to this point. To ensure the read happens at most once per type — even if {{ClipboardItem/getType()}} is called multiple times concurrently for the same MIME type — the resulting {{Promise}} is cached in [=this=]'s [=ClipboardItem/representations with resolvers=] map, keyed by a normalized type string so that `"text/html"` and `"web text/html"` are tracked independently while differently-cased spellings of the same type (e.g., `"text/HTML"` and `"text/html"`) collapse to a single entry.
+
+						1. Let |key| be |mimeType|'s [=MIME type/essence=]. If |isCustom| is true, prefix |key| with `"web "`.
+
+						1. If [=this=]'s [=ClipboardItem/representations with resolvers=][|key|] [=map/exists=], then return [=this=]'s [=ClipboardItem/representations with resolvers=][|key|].
```
EdgarChen, line 949: So if the `ClipboardItem` is created by `read()`, two `getType()` calls on it can return same promise instance. And if the `ClipboardItem` is constructed from script, `getType()` return different promise instance. All browsers currently always return different promise in either case. Is this intentional behavior changes and differences?

index.bs
```diff
@@ -1033,33 +1107,17 @@ url: https://storage.spec.whatwg.org/#obtain-a-storage-key-for-non-storage-purpo
...
-						1. Resolve |representation|'s [=representation/data=] with |systemClipboardRepresentation|'s [=system clipboard representation/data=].
-
-							Issue: It should be possible to read the data asynchronously from the system clipboard after the author calls getType, however, this set of steps implies that data will be provided at the time of read.
-						
-						1. The user agent, MAY sanitize |representation|'s [=representation/data=], unless |representation|'s [=representation/MIME type=]'s essence is "image/png", which should remain unsanitized to preserve meta data, or if it satisfies the below conditions:
-						
-							1. |representation|'s [=representation/MIME type=] is in optional [=unsanitized data types=] list.
-							
-							1. |isUnsanitized| is true.
+							Note: The OS clipboard data for this [=representation=] will not be fetched until the author calls {{ClipboardItem/getType()}} on the resulting {{ClipboardItem}} object.
```
EdgarChen, line 1112: And also the representation’s [data](https://pr-preview.s3.amazonaws.com/shwetabin/clipboard-apis/pull/248.html#representation-data) isn't used then?

index.bs
EdgarChen: Or maybe we could have a way to avoid representation's data being resolved multiple times instead, e.g.
- Have a flag that indicates the representation's data is loading from system clipboard.
- Or make representation's data is nullable, and initialize it here instead of in read().
- Or something else ..

WDYT?

index.bs
shwetabin: In ClipboardItem::getType, lazy (read()-created) items cache the resolver in representations_with_resolvers_ and return the same promise for repeated getType() of the same type; eager (author-constructed) items return a fresh .Then(...) promise each call. The cache is what guarantees the OS clipboard is read at most once per type

index.bs
shwetabin: Correct — for read()-created items the per-representation data promise isn't used; the deferred OS read populates the resolver map instead. The representation data promises drive only the author-constructed path.

index.bs
shwetabin: Removed the duplicated non-normative "When the ClipboardItem is created from read()…" sentences from the three dfns; normative setters already exist in read()

EdgarChen (approved):
Looks good to me.
I think we still need approval from @annevk as well.

annevk (approved):

